PR #75, opened 2021-09-13
Feature/1988 sign out refresh

Kyle-Herceg:
## Pull Request

Added refresh token on customize-able interval. Also rewrote inactivity timer and moved to top of component hierarchy. Inactivity timer includes general session inactivity and inactivity when checking out a record. Furthermore, added front end functionality to connect to easey-auth-api backend. Added tests for all of above.


Qb10 (changes requested):
pull down latest from develop to resolve the conflicts

PR #82, opened 2021-09-15
Feature/2015 convet axios

Kyle-Herceg:
## Pull Request
Added secure axios calls, change sign out to check user out, and bug fixes.

### DIFF
--- a/src/additional-functions/checkout.js
+++ b/src/additional-functions/checkout.js
@@ -13,8 +13,9 @@ export const checkoutAPI = (
     return mpApi
       .deleteCheckInMonitoringPlanConfiguration(monitorPlanId)
       .then((res) => {
-        console.log(res, "checked back in ");
-        if (setCheckout !== undefined) setCheckout(false, configID);
+        if (setCheckout !== undefined) {
+          setCheckout(false, configID);
+        }
         if (res === undefined) {
           console.log("error");
         }
--- a/src/additional-functions/checkout.js
+++ b/src/additional-functions/checkout.js
@@ -4,27 +4,24 @@ import * as mpApi from "../utils/api/monitoringPlansApi";
 export const checkoutAPI = (
   direction,
   configID,
-  selectedConfig,
+  monitorPlanId,
   setCheckout
 ) => {
   const user = JSON.parse(sessionStorage.getItem("cdx_user"));
 
   if (!direction) {
-    mpApi
-      .deleteCheckInMonitoringPlanConfiguration(selectedConfig.id)
+    return mpApi
+      .deleteCheckInMonitoringPlanConfiguration(monitorPlanId)
       .then((res) => {
         console.log(res, "checked back in ");
-        setCheckout(false, configID);
+        if (setCheckout !== undefined) setCheckout(false, configID);
         if (res === undefined) {
           console.log("error");
         }
       });
   } else {
-    mpApi
-      .postCheckoutMonitoringPlanConfiguration(
-        selectedConfig.id,
-        user.firstName
-      )
+    return mpApi
+      .postCheckoutMonitoringPlanConfiguration(monitorPlanId, user.userId)
       .then((res) => {
         setCheckout(true, configID);
         if (res === undefined) {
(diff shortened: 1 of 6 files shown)


src/additional-functions/checkout.js
```diff
@@ -4,27 +4,24 @@ import * as mpApi from "../utils/api/monitoringPlansApi";
...
 
   if (!direction) {
-    mpApi
-      .deleteCheckInMonitoringPlanConfiguration(selectedConfig.id)
+    return mpApi
+      .deleteCheckInMonitoringPlanConfiguration(monitorPlanId)
       .then((res) => {
         console.log(res, "checked back in ");
-        setCheckout(false, configID);
+        if (setCheckout !== undefined) setCheckout(false, configID);
```
tdavydets: should be re-written to include {} after if.

Example:
    if (condition) {
      // code here
    }

src/components/InactivityTracker/InactivityTracker.js
```diff
@@ -14,12 +13,14 @@ export const InactivityTracker = ({ openedFacilityTabs, setCheckout }) => {
   const [showInactiveModal, setShowInactiveModal] = useState(false);
 
   const isFacilityCheckedOut = () => {
-    return (
-      openedFacilityTabs.length > 0 && openedFacilityTabs[0].checkout === true
-    );
+    for (let i = 0; i < openedFacilityTabs.length; i++) {
```
tdavydets: this function can be simplified down to 1 line:

const isFacilityCheckedOut = () => {
  return (openedFacilityTabs.find((element) => element.checkout === true).length > 0);
};

src/components/InactivityTracker/InactivityTracker.js
```diff
@@ -33,8 +34,6 @@ export const InactivityTracker = ({ openedFacilityTabs, setCheckout }) => {
       console.log("Time is up!");
       resetUserInactivityTimer();
 
-      if (handler !== undefined) handler();
-
       logOut(undefined);
```
tdavydets, line 32: passing 'undefined' is not needed.  not passing anything is the same as passing 'undefined' logically.  a "default" parameter should be used in the function definition, if needed.

Example:

  const logOut = (someArgument = "defaultValue") => {
    // code here
  }

src/utils/api/easeyAuthApi.js
```diff
@@ -1,66 +1,70 @@
...
-      statusCode: e.response.data.statusCode,
-      error: e.response.data,
-    };
-  }
+    .catch((e) => {
+      throw e;
+    });
 }
 
 export async function logOut(event) {
```
tdavydets: I don't think this needs to be async.  since it doesn't return anything, nor await anything (other than what is done out of a callback [then()] function, which is its own thing), we can just treat it as a normal function

src/utils/api/easeyAuthApi.js
```diff
@@ -1,66 +1,70 @@
...
-      });
-  } catch (e) {
-    throw e;
-  }
+    .catch((e) => {
+      throw e;
+    });
 }
 
 export async function refreshToken() {
```
tdavydets: this function doesn't need to be async since it doesn't "await" anything

src/utils/api/monitoringPlansApi.js
```diff
@@ -104,18 +104,20 @@ export const getMonitoringAnalyzerRanges = async (locId, componentRecordId) => {
...
+    data: userName,
+  })
     .then((response) => response.data)
     .catch(handleError);
 };
 
 export const revertOfficialRecord = async (id) => {
-  return axios
-    .delete(`${config.services.monitorPlans.uri}/workspace/plans/${id}/revert`)
+  return secureAxios({
```
tdavydets, line 117: eventually, when we switch over to secureAxios for all transactions, we can do this at the very top of the file with the imports to avoid changing axios to "secureAxios" everywhere.

import { secureAxios as axios } from "some/location";

tdavydets (changes requested):

tdavydets (approved):

PR #128, opened 2021-10-21
Bug/1980 508 general fixes

dionolympiaCVP:
- fixing aria-live region for inactivity timer
- maintaining correct aria-label for expand facility button 
- replacing alt's with titles with context for lock & pencil icons


dionolympiaCVP:
need to add role="img" attribute to pencil and lock icons

PR #147, opened 2021-10-27
Refactored the code that loads drop down, and addressed removing the …

sanleyCVP:
…blank entry

## Pull Request

Fixed dropdown default options and related saving behavior
```


tdavydets (changes requested):
I would rename the variable from 'x'.  this is not a part of anything done for this PR, but as a rule, descriptive variable names are very important

tdavydets (approved):

PR #282, opened 2022-01-11
Bug/2265 fixing aria labels for data table title and back buttons

dionolympiaCVP:


dionolympiaCVP:
working on code smells

PR #308, opened 2022-01-31
Feature/2326 pre filter dropdowns mats method

dionolympiaCVP:


dionolympiaCVP:
- unit tests in progress
- need to create reusable function in `retrieve-dropdown-api.js` for creating the pre-filter arrays

PR #457, opened 2022-07-08
Feature/3776 remove test summary data linearity summary

cw-cvp:


addisgmichael (changes requested):
I like the reusable component for a confirmation modal. However, we have naming conventions where a directory name should match at least one JS file name and the component name. In this case, you can update the directory name to be ConfirmationModal in addition to updating the respective JS file and component names the same.

addisgmichael (approved):

PR #495, opened 2022-08-10
Feature/3746 global view protocol gas data

cw-cvp:


src/components/qaDatatablesContainer/QALinearitySummaryExpandableRows/ProtocolGasDataRows.js
```diff
@@ -0,0 +1,94 @@
...
+          />
+        </div>
+      }
+    </>
+  )
+}
+
+export default ProtocolGasDataRows
+
+const mapProtocolGasDataToRows = (data = []) => {
```
Qb10: i would move this to its own file in utils/selectors to keep align with the rest of the codebase

Qb10 (changes requested):

src/components/qaDatatablesContainer/QALinearitySummaryExpandableRows/ProtocolGasDataRows.js
```diff
@@ -0,0 +1,94 @@
...
+    'Gas Type Code',
+    'Cylinder ID',
+    'Vendor ID',
+    'Expiration Date',
+  ];
+
+  useEffect(() => {
+    const fetchProtocolGas = async () => {
+      setLoading(true)
+      const resp = await getQAProtocolGas(locationId, id)
```
addisgmichael, line 30: can you add a validation to the API response to check if the response is a success, and add a handler if not.  Also this way when you run the associated unit test, it won't complain.

src/components/qaDatatablesContainer/QALinearitySummaryExpandableRows/ProtocolGasDataRows.js
```diff
@@ -0,0 +1,94 @@
...
+import React, { useState, useMemo, useEffect } from "react";
+import { Button } from "@trussworks/react-uswds";
+
+import { getQAProtocolGas } from "../../../utils/api/qaCertificationsAPI";
+import QADataTableRender from "../../QADataTableRender/QADataTableRender";
+import { Preloader } from "@us-epa-camd/easey-design-system";
+
+const tableLabel = 'Protocol Gas'
+
+const ProtocolGasDataRows = ({
```
addisgmichael, line 12: also this component is supposed to be reusable for other test type groups being selected so better to move it under the folder qaDatatablesContainer

src/components/qaDatatablesContainer/QALinearitySummaryExpandableRows/QALinearitySummaryExpandableRows.js
```diff
@@ -1,405 +1,32 @@
...
-export { mapStateToProps };
+    <>
+      <LinearitySummaryDataRows
+        user={user}
+        nonEditable={nonEditable}
+        locationSelectValue={locationSelectValue}
+        data={data}
+      />
+      <ProtocolGasDataRows
+        user={user}
```
addisgmichael, line 23: i don't think we need a separate component to render both LinearitySummaryDataRows and ProtocolGasDataRows as you can just import and throw it in QALinearitySummaryExpandableRows

PR #511, opened 2022-08-23
Feature/2974 view add rata summary data

cw-cvp:


src/utils/selectors/QACert/TestSummary.js
```diff
@@ -147,4 +147,21 @@ export const getRataDataRecords = (data) => {
...
+}
+
+export const mapRataSummaryToRows = (data) => {
+  const records = []
+  for (const el of data) {
+    const row = {
+      id: el.id,
+      col1: el.operatingLevelCode,
+      col2: el.referenceMethodCode,
+      col3: el.apsIndicator,
```
Qb10: i think for indicator since it returns as a radio button, you need to condition check it if its either null/0 to return NO for the data table columns, or YES if it is a 1 

doublecheck with other indicators in the selector folder

Qb10 (changes requested):

addisgmichael (approved):

PR #533, opened 2022-09-09
Test coverage

cw-cvp:


addisgmichael (commented):
Please don't update UI components for test coverage PRs. it may end up breaking something and it'll be difficult to spot as we don't do functional testing for test coverage.

PR #570, opened 2022-09-26
Feature/2956 edit rata traverse data

cw-cvp:


addisgmichael (commented):
The updates look exactly similar to the PR for view and add traverse data.

addisgmichael (approved):

PR #573, opened 2022-09-29
fixed dry molecular issue and static stack pressure issue

RobertB-CVP:
Issue 1: Dry Molecular is a date pickers instead of a text box - fixed

Issue 2: Rata Flow view functionaility does not work - example was working as intended

Issue 3: Stack Static Pressure (Text) does not save - fixed

Issue 4: Edit Rata Flow does not save - fixed

### DIFF
--- a/src/utils/selectors/QACert/assert.js
+++ b/src/utils/selectors/QACert/assert.js
@@ -285,12 +285,11 @@ export const saveDataSwitch = (userInput, name, location, id, extraIdsArr) => {
     case flowRataRun:
       return qaApi
         .updateFlowRunData(
-          location,
-          id,
           extraIdsArr[0],
           extraIdsArr[1],
           extraIdsArr[2],
           extraIdsArr[3],
+          id,
           userInput.id,
           userInput
         )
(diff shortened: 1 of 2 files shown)


src/utils/selectors/QACert/assert.js
```diff
@@ -285,12 +285,11 @@ export const saveDataSwitch = (userInput, name, location, id, extraIdsArr) => {
     case flowRataRun:
       return qaApi
         .updateFlowRunData(
-          location,
-          id,
           extraIdsArr[0],
           extraIdsArr[1],
           extraIdsArr[2],
           extraIdsArr[3],
+          id,
```
addisgmichael, line 290: This update is related to editing flow rata run. is this with in the scope of ticket being worked on? I'm thinking it may break edit functionality for flow run. If not, looks good.

addisgmichael (commented):
Sorry i didn't see list of fixes above. Approved

addisgmichael (approved):

src/utils/selectors/QACert/assert.js
RobertB-CVP: This was related to an issue with editing flow rata run not saving data brought up by Moses.

PR #608, opened 2022-10-20
Feat/4106 UI implement emissions main view layout

fanwari1:
* Adds layouts for emissions within HeaderInfo
* Added MultiSelectCombobox component
* Added PillButton component, needed by MultiSelectCombobox
* Added fortawesome icon lib

### DIFF
--- a/src/components/EmissionsTab/EmissionsTab.js
+++ b/src/components/EmissionsTab/EmissionsTab.js
@@ -1,7 +1,6 @@
 import React, { useState, useEffect } from "react";
 
 import { connect } from "react-redux";
-import MonitoringPlanTabRender from "../MonitoringPlanTabRender/MonitoringPlanTabRender";
 import {
   setSectionSelectionState,
   setLocationSelectionState,
@@ -10,9 +9,10 @@ import {
 } from "../../store/actions/dynamicFacilityTab";
 import {
   convertSectionToStoreName,
-  MONITORING_PLAN_STORE_NAME,
+  EMISSIONS_STORE_NAME,
 } from "../../additional-functions/workspace-section-and-store-names";
-export const MonitoringPlanTab = ({
+import EmissionsTabRender from "../EmissionsTabRender/EmissionsTabRender";
+export const EmissionsTab = ({
   resetTimer,
   setExpired,
   resetTimerFlag,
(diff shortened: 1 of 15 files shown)


src/components/EmissionsTab/EmissionsTab.js
```diff
@@ -0,0 +1,126 @@
...
+
+  }
+  const [ currentTabIndex, setCurrentTabIndex ] = useState(getCurrentTabIndex());
+
+  useEffect(()=>{
+    // setCurrentTab(getCurrentTab())
+    setCurrentTabIndex(getCurrentTabIndex())
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  },[selectedConfig,checkout,currentTabIndex]);
```
eglove: effects aren't needed for computed values. This is going to set on every render regardless. Using an effect here makes it run twice. Memo is not necessary either because the calculation isn't expensive enough for it to be faster. (Memo runs a deep comparison that would slow this down.)

All that's needed is const currentTabIndex = getCurrentTabIndex()

src/components/EmissionsTab/EmissionsTab.js
```diff
@@ -0,0 +1,126 @@
...
+
+  const [locationSelect, setLocationSelect] = useState(
+    getCurrentTab().location
+  );
+
+  useEffect(() => {
+    setLocation(locationSelect, title, workspaceSection);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [locationSelect]);
```
eglove, line 50: We'll not bother with the fact that there's a state setter sitting somewhere in Redux. I don't think there's a comfortable way to fix this.

src/components/MultiSelectCombobox/MultiSelectCombobox.js
```diff
@@ -0,0 +1,256 @@
...
+  const [showListBox, setShowListBox] = useState(false);
+  const [selectedItems, setSelectedItems] = useState([]);
+  const selectedItemsRef = useRef(selectedItems);
+  const [stillMounted, setStillMounted] = useState(true);
+
+  useEffect(() => {
+    populateSelectedItems();
+    return () => setStillMounted(false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
```
eglove: populateSelectedItems is going to be recreated on every render. Even if it doesn't change, a new one exists on every instance of the component.

There's no reason for this eslint ignore, it should contain [populateSelectedItems] and the effect can be moved below all the declared functions.

src/components/MultiSelectCombobox/MultiSelectCombobox.js
```diff
@@ -0,0 +1,256 @@
...
+        const multiSelectComboboxDiv = document.getElementById(
+          `multi-select-combobox-${entity}`
+        );
+        if (
+          multiSelectComboboxDiv &&
+          !multiSelectComboboxDiv.contains(e.target)
+        ) {
+          setShowListBox(false);
+        }
+      })}
```
eglove: Double check that this isn't a memory leak, if this event listener is being added on every render but not removed it was cause a major performance drag.

Event listeners should be added in effects and removed in their returns.

eglove (changes requested):

src/components/MultiSelectCombobox/MultiSelectCombobox.js
fanwari1: Updated

src/components/MultiSelectCombobox/MultiSelectCombobox.js
fanwari1: It does indeed and code has been updated to clean up the event.

PR #683, opened 2022-11-25
feature / 4523 - Hide 'Revert to Official' Button from QA UI

billmiller13:
Adds a 'false' condition to hide the Revert to Official Button temporarily, until the button functionality is implemented.


subrataroyCVP (approved):
LGTM

PR #729, opened 2022-12-09
Made cycle time summary data table expandable. Added edit and remove

jordanfontainecvp:
made cycle time summary data table expandable. Added edit and remove functionality


addisgmichael (approved):
LGTM

PR #771, opened 2022-12-23
Fix Bug in HeaderInfo when No currentTab

billmiller13:
Recent changes to HeaderInfo introduced a bug when there is no "currentTab", causing a couple of unit tests to fail


fanwari1 (approved):
lgtm

PR #775, opened 2022-12-27
Fixed layout of apply filter button and reporting period

fanwari1:
Made view template dropdown 

Apply filter button and reporting period should be moved in line with other drop downs in 1200 and 1400 view


sdhawan16 (approved):
lgtm

PR #859, opened 2023-02-05
Feat/4587 create error suppression table layout

fanwari1:
## Pull Request

Adds the table to see error suppression data

A few values have been hardcoded into the api call which will later come from the filters


sdhawan16 (approved):
lgtm

PR #877, opened 2023-02-09
Feat/4684 implement apply filters button functionality in error suppression module

fanwari1:
## Pull Request

This PR removes the hard coded values for the Error Suppression GET api call and implements the apply filter button which useses actual values from the selected filters.


sdhawan16 (approved):
LGTM

PR #886, opened 2023-02-10
Bugfix/3672 modal errors

cw-cvp:


addisgmichael (approved):
LGTM. Pls add test coverage as it seems low on new code. Can be done after merging this so that it's tested in time.

PR #897, opened 2023-02-16
Feature/4798 api communication error

RobertB-CVP:
Location: API Communication Errors
Issue:

On the API Communication errors, the contrast is lower than 4.5:1
API communication error is not in Aria live


jordanfontainecvp (approved):
lgtm

PR #899, opened 2023-02-16
Feature/4796 table header focus

RobertB-CVP:
Locations: Any tables/grid in ECMPS
Issues:

Table headers do not have an up/down arrow visible on focus only on mouseover (Select Configurations Screen: Tab to the Header and the Sort arrow needs to be visible)
The fix needs to be implemented for both Logged-in and Global-view and across ECMPS (MP, QA, Emissions, Export etc wherever we have tables with Sort)


fanwari1 (approved):
LGTM!

PR #917, opened 2023-02-23
bug/4783-508-focus-box-visibility-on-table-column-headers

fanwari1:
## Pull Request

Fixed an issue of the focus from table column disappearing which clicking on the table to sort. This change was made in the global table css file so it should apply to all tables.


sdhawan16 (approved):

sdhawan16 (approved):
lgtm

PR #943, opened 2023-03-07
Feature/4871 sustain active tab

RobertB-CVP:
Issue:
When a user opens multiple tabs and is actively working in one tab -> upon navigation to another module or option from the left-hand menu, the focus automatically goes to the tab that was opened last even when it might not be the one that the user was working in

Scenarios to test:

When user opens multiple tabs, and navigates away to another module then the user should be on the last tab/Configuration opened
When a user opens multiple tabs, clicks on any one of them, and navigates to a different module, then the focus should be retained on the tab that the user was on.


addisgmichael (commented):
Alternatively you can maintain currentTabIndex in Redux (global state) and pull it in any component you like.

billmiller13 (approved):

PR #956, opened 2023-03-15
Fixed error where null dates would cause crashing

RobertB-CVP:
Bug: Expanding Cycle Time summary record crashes system

Steps to Reproduce
Log in to ecmps
Check out facility 3776 Glem Lyn (51)
Select cycle time summary as a test group
Exanpd the second row of test data
Expand the first row of cycle time summary data
The system crashes

Second row of test data


billmiller13 (approved):
lgtm

PR #959, opened 2023-03-16
Bug/4954 error message inconsistency

cw-cvp:
Any error message being displayed on the ECMPS screens need to be formatted in a readable manner. For instance, a message where there are multiple errors combined and displayed in one go needs to be split to make it readable


RobertB-CVP (dismissed):

RobertB-CVP (dismissed):

RobertB-CVP (approved):
Nice coverage

PR #969, opened 2023-03-20
Feature/4967 mp update date format

cw-cvp:
1. Update the MP Date format as below where we have the Date and Time displayed as one column in the grid
Format: YYYY/MM/DD H
Note: If end date is not available and hours/mins are, it should just leave it blank

2. Update the MP Date format as below:
Format: YYYY/MM/DD


addisgmichael (commented):
Can you pls make sure when an hour/min is provided it has a 2 digit pad. eg. 2 should be 02

addisgmichael (approved):

PR #983, opened 2023-03-24
bug/4987

cw-cvp:
The system appears to be crashing when a user imports a file that has a test number with three dashes (ex."testNumber": "EPA-130-6001-Janell")


jordanfontainecvp (approved):
lgtm

PR #1017, opened 2023-04-12
Bug/5036 UI Links in the Home Page

sdhawan16:
## Pull Request

```

- Added links on home page for qa certification and emissions 
- modified URL for emissions global
- when page is reloaded from buttons on home page tab order is reset 

```


fanwari1 (approved):
lgtm

PR #1086, opened 2023-05-08
Update HeaderInfo.js

cw-cvp:
Emissions module is not refreshing properly when a user is logged in and imports a file

 Reporting period, location, and view template should be refreshed based on the imported file

 Whatever reporting period is imported should be automatically added to the reporting period drop down to refresh the view template

### DIFF
--- a/src/components/HeaderInfo/HeaderInfo.js
+++ b/src/components/HeaderInfo/HeaderInfo.js
@@ -234,7 +234,7 @@ export const HeaderInfo = ({
   const MAX_REPORTING_PERIODS_ERROR_MSG =
     "You can only select a maximum of four reporting periods";
 
-  const reportingPeriods = useMemo(
+  let reportingPeriods = useMemo(
     () =>
       getReportingPeriods().map((reportingPeriod, index) => {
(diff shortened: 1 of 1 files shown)


src/components/HeaderInfo/HeaderInfo.js
```diff
@@ -227,26 +227,26 @@ export const HeaderInfo = ({
...
-      : "Test";
+        ? "Emissions"
+        : "Test";
 
   const MAX_REPORTING_PERIODS = 4;
   const MAX_REPORTING_PERIODS_ERROR_MSG =
     "You can only select a maximum of four reporting periods";
 
-  let reportingPeriods = useMemo(
+  const reportingPeriods = useMemo(
```
fanwari1: Can you lave this as let instead of const. This variables looks like it gets reassigned else where in the code from what I'm seeing inside the handleSelectedReportingPeriod function. If it is const then it'll throw an error when that function runs.

cw-cvp: Reverted to const.

fanwari1 (approved):

PR #1125, opened 2023-06-19
fix/5212 - Keep Selected Report Periods on Emissions Checkout/In

billmiller13:
When the selected Reporting Periods was changed and then the Checkout or Check Back In button was clicked (without first clicking the "Apply Filters" Button), the Reporting Periods combo select box was reverting to the previous selections.  This change locks in the selections that are made before Check Out/In is clicked so the re-render doesn't restore the old selections.


sdhawan16 (approved):
lgtm

PR #1144, opened 2023-06-28
Feature/5221 create view em submission popout layout

cw-cvp:


fanwari1:
@cw-cvp I ran this and I'm not actually seeing any values:
image
 Also we made some changes so that certain fields that are in the spec aren't actually needed anymore like Analyst and Representative so those should probably be removed and there are a few others as well so you can look at the table and see what columns we show and only have those in the popout.

Also looking at the wireframe, it looks like there's a title there, plus a close button at the bottom. Also there's a gray background so that users can't click on anything else while the popout is open.

subrataroyCVP:
Seeing similar like @fanwari1 . No values in the modal.

PR #1162, opened 2023-07-07
Feature/5238 qa cert maintenance table

cw-cvp:


src/components/AdminMaintenance/FilterFormAdmin/FilterFormAdmin.js
```diff
@@ -233,7 +304,7 @@ const FilterFormAdmin = ({
               />
             </div>
           </Grid>
-          {section === QA_CERT_DATA_MAINTENANCE_STORE_NAME ? (
+          {section === QA_CERT_DATA_MAINTENANCE_STORE_NAME &&
```
fanwari1: When it comes to jsx, it's best practice to use terneries instead of &&. You can look here for an explanation:
https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx

src/components/AdminMaintenance/AdminMaintenance.js
```diff
@@ -85,16 +86,31 @@ export const AdminMaintenance = ({ section }) => {
...
+          typeSelection={qaMaintenanceTypeSelection}
+        />
+      }
     </div>
   );
-};
\ No newline at end of file
+};
+
+export const testSummaryLabel = 'Test Summary'
```
fanwari1: Can we keep these labels inside FilterFormAdmin.js file? They don't really have anything to do with AdminMaintenance component

src/utils/api/camdServices.js
```diff
@@ -58,3 +58,28 @@ export const sendSupportEmail = async (payload) => {
     throw new Error(error);
   }
 };
+
+const getMaintenanceRecords = async (orisCode, unitStack, endpoint) => {
```
fanwari1: Can we move these to AdminMaintenanceApi.js? That's where we are keeping all the admin maintenance endpoints

src/components/AdminMaintenance/FilterFormAdmin/FilterFormAdmin.js
cw-cvp: switched conditional rendering from && to ternary

fanwari1 (changes requested):

src/components/AdminMaintenance/AdminMaintenance.js
cw-cvp: label variables have been moved into and exported from FilterFormAdmin.js

src/utils/api/camdServices.js
cw-cvp: api calls have been moved into AdminManagementApi.js

subrataroyCVP:
Why `Cert Events` and `TEE` showing propertyNames instead empty value?

image

If there is no data we should show something like this. ⬇️

image

cw-cvp:
> Why `Cert Events` and `TEE` showing propertyNames instead empty value?
> 
> image
> 
> If there is no data we should show something like this. ⬇️
> 
> image

Currently we are displaying dummy data. I believe a later ticket will connect the api to fetch real data.

subrataroyCVP:
image

This should be something like this ⬇️

image

Example:
image

subrataroyCVP:
> > Why `Cert Events` and `TEE` showing propertyNames instead empty value?
> > image
> > If there is no data we should show something like this. ⬇️
> > image
> 
> Currently we are displaying dummy data. I believe a later ticket will connect the api to fetch real data.

Then put the place holder for `No data available.` As per other table.

When we will connect the APIs for other table we will display data.

subrataroyCVP:
Apology. I thought there are data for test-summary. Could you revert the dummy data, so table headers are shown.

Sorry for that.

subrataroyCVP (approved):
LGTM

fanwari1 (approved):

PR #1196, opened 2023-07-26
Feature/5359 test files group 15

cw-cvp:


addisgmichael (changes requested):
Standardize
- use reusable renderer from src/mocks/render.js
- use constants.js and functions.js under src/mocks to populate mock data

cw-cvp:
> Standardize
> 
> * use reusable renderer from src/mocks/render.js
> * use constants.js and functions.js under src/mocks to populate mock data

changes made

src/components/export/Export/Export.test.js
```diff
@@ -88,20 +89,19 @@ initialState.openedFacilityTabs.export = [
...
+  title: "Barry (1, 2, CS0AAN)",
 };
 
-test("testing Export component renders with no errors", async () => {
-  let { container } = await waitForElement(() => componentRenderer());
-  expect(container).toBeDefined();
-});
+test('renders Export', async () => {
+  await act(() => {
+    render(
```
addisgmichael: render function is already wrapped with act and a provider if you pass in store. you can just call it this way
 await render(<Export {...props}/>, store)

addisgmichael (approved):

PR #1223, opened 2023-08-17
fix/5580 - Save Unit Control 'Original Code' as String

billmiller13:
To match API Spec, converts numeric 0/1 for Original Code (radio value) to "0" or "1" when creating a new Unit Control.  This fix was already made for Unit Control updates.


subrataroyCVP (approved):
LGTM

PR #1231, opened 2023-08-23
Bug/5568 mats bulk file import local environment

fanwari1:
## Pull Request

This PR adds our localized version of FileInput with a bug fix where the app would crash due to not being able to load the file properly on our local system.


src/components/ErrorSuppression/ErrorSuppression.js
cw-cvp, line 1: Should this FileInput import be from the local implementation?

cw-cvp, line 1: Should this FileInput import be from the local version?

fanwari1: pushed up changes

cw-cvp (approved):

PR #1305, opened 2023-10-06
update table id

jeffreyaggrey:


subrataroyCVP (changes requested):
As per the issue its saying Record ID . Please change it to Record ID.

That's what is done in Error Suppression.

Screenshot 2023-10-06 135528

subrataroyCVP (approved):

PR #1318, opened 2023-10-12
Feature/5746 config view

Qb10:


subrataroyCVP (commented):
LGTM. 

Please check with @jason if we want to add a tooltip to see the whole text when its hidden.

image

subrataroyCVP (dismissed):

jwhitehead77:
@Qb10 please add hover text showing the entire config value since we are cutting it off

subrataroyCVP (approved):
LGTM.

@jwhitehead77 It is showing just a basic tooltip. Actually not only this its everywhere. I guess we need to use uswds tooltip.

https://designsystem.digital.gov/components/tooltip/#tooltip-preview

PR #1337, opened 2023-10-19
feat/adding-submission-type-and-reason-columns-to-em-admin

fanwari1:
## Pull Request

Removed Facility, State, and Locations columns from em submission admin module table

Added Submission Type column


subrataroyCVP (approved):
LGTM

PR #1370, opened 2023-11-02
Bug/5866 mp component view duplicate

cw-cvp:


src/components/datatablesContainer/DataTableSystems/DataTableSystems.module.scss
```diff
@@ -0,0 +1,3 @@
+.returnLink {
```
addisgmichael: pls change file name to match component file name
class selector has to be wrapped with a namespace so that this style is only applicable to the associated component. See src\components\DataTableRender\DataTableRender.scss

cw-cvp: Changed file name

addisgmichael (approved):

PR #1393, opened 2024-01-12
The Sponsor role or Initial Authorizer need to have the ability to su…

### LINKED ISSUE #5994: The "Sponsor" role or "Initial Authorizer" need to have the ability to submit- UI
The "Sponsor" role or "Initial Authorizer" need to have the ability to submit. Code in easey-auth-api/src/sign/Sign.service.ts (authenticate method) continues to limit submissions to users with the "Submitter" role.  

 

<<<>>> 

if (!user.roles.includes('Submitter')) {
   throw new EaseyException(
     new Error('This requires the Submitter role'),
     HttpStatus.BAD_REQUEST,
     {
       responseObject: 'This requires the Submitter role',
     },
   );
 }

<<<>>> 

 

This logic needs to be expanded to also allow users with the "Sponsor" role or "Initial Authorizer" role to make submissions. If there is already a zenhub ticket for this issue, then please send it along.

shahbazkhan1999:


spetros-do (approved):
looks good.

PR #1394, opened 2024-01-30
Max-age change to min default;

spetros-do:
Max-age needs to be set to min the default of 1 year as opposed to current setting had to 6 months.
--


ergjustin (approved):
looks good to me.

PR #1408, opened 2024-03-22
fix:componentId and moniteringSystemId read only for submited QA

### LINKED ISSUE #6060: Make component and system ID fields read-only for submitted QA tests
In the workspace, users are able to change component and system IDs for already submitted QA tests.

These fields should be read-only for already submitted tests. The current practice is to first delete these tests if they were submitted with incorrect component/system IDs. This ensures related supplemental data is also deleted and invalid evaluation results are not received.

ergjustin:
## Ticket
[Make component and system ID fields read-only for submitted QA tests](https://github.com/US-EPA-CAMD/easey-ui/issues/6060)

## Changes
`componentId` and `monitoringSystemId` read only for submited QA test data

### DIFF
--- a/src/components/ModalDetails/ModalDetails.js
+++ b/src/components/ModalDetails/ModalDetails.js
@@ -29,6 +29,7 @@ const ModalDetails = ({
   setMainDropdownChange,
   mainDropdownChange,
   setDisableExitBtnStatus,
+  disableColoums
 }) => {
   // fixes resizing issue with calendar date picker along with help in CSS file
   const containerStyle = {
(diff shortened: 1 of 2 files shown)


src/components/ModalDetails/ModalDetails.js
```diff
@@ -29,6 +29,7 @@ const ModalDetails = ({
   setMainDropdownChange,
   mainDropdownChange,
   setDisableExitBtnStatus,
+  disableColoums
```
djw4erg: Although it does not negatively affect the change, "Coloums" in the field name should be "Columns" and I would argue for naming it "disabledColumns".  Minimally, correct the spelling.

djw4erg (changes requested):

djw4erg (approved):

PR #1413, opened 2024-04-02
fix:second normal indicator radio button to dropdown

### LINKED ISSUE #6059: Add null option for Load - Second Normal Indicator
In the monitoring plan load records, there is not an option for null for the second normal indicator. This value is not required for non load-based or peaking units or stacks linked to peaking units.

Yes and No are the only options. A null option should be added.

If a peaking unit for example makes a change to the load record (max load value, lower/upper boundary, etc.) and saves, second normal indicator is set to No. A subsequent evaluation will then generate an unnecessary LOAD-21-A non-critical error.

ergjustin:
## Ticket
[Add null option for Load - Second Normal Indicator](https://github.com/US-EPA-CAMD/easey-ui/issues/6059)

## Change

- Second Normal Indicator radio button changed to dropdown


maxdiebold-erg (approved):
Looks good!

PR #1414, opened 2024-04-05
fix:remove system component have length condition

### LINKED ISSUE #5998: Systems - add component popup window when adding a new system
After adding a new system, the first time the add new component popup window appears, the dropdown selection box is null. The user is forced to create a new component to continue.

ergjustin:
## Ticket
[Systems - add component popup window when adding a new system - Dropdown Issue](https://github.com/US-EPA-CAMD/easey-ui/issues/5998)

## Change

- Removed the if condition of `sysComps.length >= 1`


maxdiebold-erg (changes requested):
It looks like the filtering is correct, but the updated state is not rendering properly:

components

I believe that removing the conditional wrapping the block starting at line 63 should fix the remaining issue.

maxdiebold-erg (approved):
Looks good!

PR #1416, opened 2024-04-10
fix:update reporting periods state after import historical

### LINKED ISSUE #6049: Reporting Period Dropdown Issue in Emissions Module
When emissions data is imported from historical, the reporting period dropdown will change to the correct reporting period (if 2023 Q3 emissions data is imported, 2023 Q3 will be selected from the dropdown automatically). When initially selecting a view after import or switching between units in the configuration, the reporting period dropdown will select both the reporting period of the imported data and the current reporting period. The user will then have to go back into the dropdown to remove the current reporting period selection to view the imported data.

ergjustin:
## Ticket
[Reporting Period Dropdown Issue in Emissions Module](https://github.com/US-EPA-CAMD/easey-ui/issues/6049)

## Changes

- update `reportingPeriods` state after import historical

### DIFF
--- a/src/components/HeaderInfo/HeaderInfo.js
+++ b/src/components/HeaderInfo/HeaderInfo.js
@@ -931,6 +931,7 @@ export const HeaderInfo = ({
     });
 
     setSelectedReportingPeriods([importedReportingPeriod])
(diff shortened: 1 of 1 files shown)


maxdiebold-erg (approved):
Looks good.

PR #1432, opened 2024-05-28
Removed the statement steps, removed double scrollbars (retained hori…

### LINKED ISSUE #5830: Sign & Submit Certification Statements Step
During the Sign & Submit process when you get to the Certification Statements there are 2 vertical scrollbars and they do not work well together. When you expand the cert statement it goes below the fold and you have to scroll to see it. Should at least auto scroll to show the cert statement when expanded. Even better would be to change the modal to only show what is needed at each step so that scrolling is not necessary.

Also update the certify button on the submission dialog to initiate the submission and eliminate the redundant submission page that appears after certification. Change "Sign and Submit" to "Sign and Submit selected files" on the page where you are selecting files for submission.

image

yonatan-dp:
…zontal scrollbar), changed certify label to "Sign and Submit selected files"

## Pull Request

### DIFF
--- a/src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
+++ b/src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
@@ -78,6 +78,7 @@ export const EvaluateAndSubmit = ({
   const modalMessage = useRef("");
 
   const [finalSubmitStage, setFinalSubmitStage] = useState(false);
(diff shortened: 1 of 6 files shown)


ergjustin (dismissed):
Looks good to me

spetros-do (approved):
All good.  Reflecting the iCAM way forward.

PR #1437, opened 2024-06-05
Feat/#6245 auto logout

### LINKED ISSUE #6245: Automatic logout after period of inactivity (AC-2(5))
NIST 800-53 Rev 5, AC-2(5) Requires that users log out when time period of inactivity exceeds thirty (30) minutes.

**Implementation Notes:**

Automatically log the user out after a period of 30 minutes of inactivity.  The current implementation warns the user when they are approaching the inactivity limit and log's them out if they do not take action to extend the session.  This is fine.  We just need to insure the limit is set to 30 minutes.

yonatan-dp:
## Pull Request


spetros-do (approved):
Looks good.

PR #1440, opened 2024-06-06
Added a check where if there is no window status for an EM, users sho…

### LINKED ISSUE #6189: Able to attempt a submission for an emissions file with no window
On the Submit screen, users are able to select an emissions file for submittal when there is not a submission window associated with that quarter.

The Submission Window column displays "None". In this situation, the record should not be selectable.

This row should not be selectable and the user should not be able to continue with Sign & Submit.

image

yonatan-dp:
Added a check where if there is no window status for an EM, users should not be allowed to submit.


spetros-do (approved):
Looks good.

PR #1447, opened 2024-06-17
Fix/#6061 qa events

### LINKED ISSUE #6061: Updates for already submitted QA Cert Events
Users have the ability to remove already submitted QA Certification Events from their workspace. This could cause invalid evaluation results. Users are also able to modify all of the fields for already submitted events. This should not be allowed. Supplemental data that is used by the check engine exists based on QA Certification Events.

- Remove the ability to delete already submitted QA Certification Events from the workspace. Events that need to be deleted or modified should first be handled using the Admin tool. This also ensures that related supplemental data is deleted when appropriate and support staff can determine if modifications to events will require EM resubmissions.
-  For already submitted QA Certification Events, the only fields that should be editable are Completion of Tests Date and Completion of Tests Hour.
- For a new QA Cert Event that has been saved but not submitted in 1.0, all fields except Completion of Tests Date and Completion of Tests Hour also become read-only. This may not need to necessarily happen in 2.0 so long as changing these fields requires a re-evaluation of the event and any dependent data.

yonatan-dp:
## Pull Request

### DIFF
--- a/src/components/ModalDetails/ModalDetails.js
+++ b/src/components/ModalDetails/ModalDetails.js
@@ -29,7 +29,10 @@ const ModalDetails = ({
   setMainDropdownChange,
   mainDropdownChange,
   setDisableExitBtnStatus,
-  disabledColumns
+  disabledColumns,
+  disableSubmittedQaEventFields,
+  editableSubmittedQaEventFields
+
 }) => {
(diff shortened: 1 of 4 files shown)


ergjustin (approved):
Looks good to me

PR #1451, opened 2024-06-20
feat:add force reEvaluation option in evaluation page

### LINKED ISSUE #6057: Add Force Re-Evaluation option
On the Evaluate screen, add an option to force re-evaluation of MP, QA, and EM data.

There is a check that requires the evaluation of MP's at least once a calendar year. There will also be times when there are plant/unit changes in CBS that are logged to INVENTORY_STATUS_LOG and subsequently require MP's to be re-evaluated. 

Users may not know the MP needs evaluation flag is set but QA/EM evaluations will indicate that the MP needs evaluation.

It would be useful for users to just be able to re-evaluate all affected data using this flag at once.

Have also come across instances in 2.0 testing where QA and EM evaluations indicate the MP needs to be evaluated even though the MP status is Passed, Updated on Host, and the Monitoring Plan cannot be selected for evaluation. The user either needs to make a temporary change to the MP or export/re-import the MP to trigger the evaluation.

ergjustin:
## Ticket
[Add Force Re-Evaluation option](https://github.com/US-EPA-CAMD/easey-ui/issues/6057)

## Changes

- Added force Re-Evaluation option in evaluation page

### DIFF
--- a/src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
+++ b/src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
@@ -79,6 +79,7 @@ export const EvaluateAndSubmit = ({
 
   const [finalSubmitStage, setFinalSubmitStage] = useState(false);
   const [shouldAutoSubmit, setShouldAutoSubmit] = useState(false);
+  const [isForceReEvaluation, setIsForceReEvaluation] = useState(false);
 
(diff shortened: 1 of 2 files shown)


usmanahmederg (approved):

usmanahmederg:
Looks good to me.

PR #1476, opened 2024-08-27
Update revert popup language

### LINKED ISSUE #6311: Requirements: Update language in pop-up when reverting to official record
After a user has imported data under the Monitoring Plan section after logging in and checking out a configuration, the user has the option to click on "Revert to Official Record". When a user clicks on this button, a pop-up warns the user that reverting to the official record will undo all saved and unsaved changes (see below). However, it does not specify that this impacts not just the present Test Data section, but all of the QA & Cert and Emissions data. Text should be updated to specify exactly what data will be reverted back - not just a general statement. This way, users know exactly what will be impacted by reverting to the official record.

Image

## Acceptance Tests

1. The following message is returned when the user clicks the "Revert to Official Record" button in the Monitoring Plan screen for a checked out configuration:

     Reverting to Official Record in the Monitoring Plan module will undo all saved and unsaved changes from the workspace. Any 
     modified QA and Emissions data will also be purged from the workspace. This is not recoverable. If you want to maintain changes in a 
     module, then export the file(s) before reverting to the official record. Do you want to continue?

ergjustin:
## Ticket
[Update language in pop-up when reverting to official record](https://github.com/US-EPA-CAMD/easey-ui/issues/6311)

## Change
Updated  reverting pop-up text.

### DIFF
--- a/src/components/HeaderInfo/HeaderInfo.js
+++ b/src/components/HeaderInfo/HeaderInfo.js
@@ -1139,7 +1139,7 @@ export const HeaderInfo = ({
           children={
             <div>
               {
-                "Reverting to Official Record will undo all saved and unsaved changes. This is not recoverable. Do you want to continue?"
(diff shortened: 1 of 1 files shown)


maxdiebold-erg (approved):
Good to go, the popup matches the text in the acceptance test.

PR #1477, opened 2024-08-28
fix  mobile menu toggle and update css

### LINKED ISSUE #6367: Bug: Mobile menu toggle button doesn't show menu
The hamburger icon in the header of the mobile view (indicated below) should open a menu when clicked, but currently does nothing.

Image

ergjustin:
## Ticket
[Mobile menu toggle button doesn't show menu](https://github.com/US-EPA-CAMD/easey-ui/issues/6367)

## Changes

- Enable mobile menu and update css

### DIFF
--- a/src/components/SubHeader/SubHeader.js
+++ b/src/components/SubHeader/SubHeader.js
@@ -56,7 +56,7 @@ export const SubHeader = ({ user, setCurrentLink }) => {
       label: (
         <span className="margin-right-1 text-no-wrap">
           <span
-            className="margin-right-20 menu-item-separator"
+            className="margin-right-20 menu-item-separator display-none desktop-lg:display-inline-block"
             style={{ color: "#365b8f" }}
           >
             |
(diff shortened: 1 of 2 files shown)


maxdiebold-erg (changes requested):
That fixes the bug, great! It looks like you fixed some other styles, too.

It also looks like the +/- icons could use a bit of adjustment, they're the same color as the background and a bit off center. I think lines 195-197 of `_uswds-overrides.scss` should be changed to:
```diff
  .usa-nav__primary button span::after {
-    background-color: $epa-white !important;
+    top: calc(100% - 1.25rem - 0.125rem - 6px);
  }
```

maxdiebold-erg:
Thanks for making those changes!

I'm surprised there aren't links to the various modules in that menu, it looks like in the mobile layout they can only be accessed from the home page, and there's no home button that I can find. I think that is an issue for a separate ticket, this looks good.

maxdiebold-erg (approved):

PR #1479, opened 2024-08-30
Fix/6366 disallow checkout

### LINKED ISSUE #6366: Don't allow a configuration to be checked out if a submission or evaluation is in progress
There is currently logic in place to prevent a user from checking a Monitor Plan _in_ if there is a submission or evaluation in progress for that plan (see `easey-monitor-plan-api` -> `user-check-out.service.ts` -> `checkInConfiguration`). However, there is no logic in place to prevent the user from checking that plan _out_ if a submission or evaluation is underway. We should duplicate the aforementioned logic to the `checkOutConfiguration` function.

More importantly, the procedure `camdecmpswks.user_session_maintenance()` deletes any row from the _camdecmpswks.user_check_out_ table with a `last_activity` value over 20 minutes old, without checking submissions or evaluations. This procedure is called by a cron job every minute. This procedure should be updated to use the same logic as the `checkInConfiguration` function.

maxdiebold-erg:
## Related Issues

- [#6366](https://github.com/US-EPA-CAMD/easey-ui/issues/6366)

## Main Changes

- Fixed the `postCheckoutMonitoringPlanConfiguration`: The `handleError` function was being shadowed by a parameter, so errors would not be handled correctly.
- Fixed some calls to the same function that were supplying incorrect arguments.

### DIFF
--- a/src/additional-functions/checkout.js
+++ b/src/additional-functions/checkout.js
@@ -22,7 +22,7 @@ export const checkoutAPI = (
       });
   } else {
     return mpApi
-      .postCheckoutMonitoringPlanConfiguration(monitorPlanId, user.userId)
+      .postCheckoutMonitoringPlanConfiguration(monitorPlanId)
       .then((res) => {
         if (setCheckout !== undefined) {
           setCheckout(true, configID, MONITORING_PLAN_STORE_NAME);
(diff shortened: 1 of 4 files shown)


ergjustin (approved):
Looks good to me!
When I try to checkout Barry (8) I was able to see the error message.

PR #1484, opened 2024-09-10
6347 Issues adding/linking a new component belonging to multiple systems

### LINKED ISSUE #6347: Issues adding/ linking a new component belonging to multiple systems
When adding/ linking a new component that belongs to multiple systems (ex. CO2, O2, DAHS, PLC, etc.), duplicate component records are added to the camdecmpswks.component table.

Ex. ORIS 136, 2. Component ID 036 (CO2) is linked to both the NOX system (system ID 204) and the CO2 system (system ID 205). When this component is ended in each system and a new one is added and linked to each system, the evaluation returns a Critical 1 COMPON-26-A error indicating the new component has not been linked to a system.

A record is getting created in camdecmpswks.component when the component is first created (correct) and when the component is linked to the second system (not correct). This is causing the COMPON-26-A error. 

This also affects the json export. The export will contain duplicate component records for the newly added component.

### Acceptance Tests: 

1. When a new component is added in the UI, a record is created in camdecmpswks.component.
2. A new record is not created in camdecmpswks.component when that component is linked to another system. 
3. Evaluating the MP changes does not result in any unexpected errors.
4. The json export only contains one component record for the newly added component (with appropriate analyzer range data where applicable).
5. Revert to official, import the json, and evaluate the MP again. No unexpected errors are returned.
6. Confirm again that only one record was created in camdecmpswks.component.
7. Verify the changes still look correct in the UI following the import.

usmanahmederg:
Ticket
https://github.com/US-EPA-CAMD/easey-ui/issues/6347

Changes
Fixed adding/linking an existing component to multiple system creating duplicated components

### DIFF
--- a/src/components/datatablesContainer/DataTableSystems/DataTableSystems.js
+++ b/src/components/datatablesContainer/DataTableSystems/DataTableSystems.js
@@ -627,13 +627,15 @@ export const DataTableSystems = ({
     try {
       let resp;
       let response;
-        resp = await mpApi
-        .createComponents(
-          userInput,
-          selectedSystem.locationId,
-          selectedSystem.id
-        )
-        .catch((error) => console.log("createComponents failed", error));
+        if (!addExistingComponentFlag) {
+          resp = await mpApi
+          .createComponents(
+            userInput,
+            selectedSystem.locationId,
+            selectedSystem.id
+          )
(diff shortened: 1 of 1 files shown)


ergjustin (approved):
Code looks good to me!

PR #1491, opened 2024-09-25
Feature/#6369 display message for responsibility free users

### LINKED ISSUE #6369: Give user a message when they have no responsibilities
In the test environment, users with the Initial Authorizer (e.g., mibarra-ia) role are able to login to ECMPS 2.0 with the CBSI Stage bypass enabled (screenshot 1).  When no facilities are returned for ECMPS 2.0 users, these users will receive the following message:

   You have no facilities for which you are responsible. Please login to CBS to review and add facilities for which you are responsible.

**Screenshot 1**
Image

EvanSun220:
## Pull Request

```

- Added related logic in component AboutHome to display warning message on the ECMPS home page, to remind user with IA role and no responsibility assigned about login to CBS to review and add facilities.

```

### DIFF
--- a/src/components/AboutHome/AboutHome.js
+++ b/src/components/AboutHome/AboutHome.js
@@ -11,6 +11,7 @@ import axios from "axios";
 import { handleResponse, handleError } from "../../utils/api/apiUtils";
 import "./AboutHome.scss";
 import {getLoginState} from "../../utils/api/easeyAuthApi";
+import { displayAppWarning } from "../../additional-functions/app-error";
 
 const getContent = async (path) => {
   const url = `${config.services.content.uri}${path}`;
@@ -28,6 +29,8 @@ const AboutHome = ({ user, setCurrentLink }) => {
   const [whatIsNewContent, setWhatIsNewContent] = useState();
   const [monitorPlanContent, setMonitorPlanContent] = useState();
   const [qaCertificationContent, setQACertificationContent] = useState();
+
+  const [warningMessage, setWarningMessage] = useState(null);
 
   useEffect(() => {
     document.title = "ECMPS Home";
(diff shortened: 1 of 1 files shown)


src/components/AboutHome/AboutHome.js
```diff
@@ -29,6 +30,8 @@ const AboutHome = ({ user, setCurrentLink }) => {
   const [monitorPlanContent, setMonitorPlanContent] = useState();
   const [qaCertificationContent, setQACertificationContent] = useState();
 
+  const [warningMessage, setWarningMessage] = useState(null);
```
yonatan-dp: This state definition is not being used anywhere in a meaningful way (we can remove this). See comment below.

src/components/AboutHome/AboutHome.js
```diff
@@ -49,6 +52,21 @@ const AboutHome = ({ user, setCurrentLink }) => {
...
+    //get the current user
+    const cdxUser = JSON.parse(localStorage.getItem("ecmps_user"));
+
+    //check if the user has initial authorizer role with no responsibilities
+    if (cdxUser?.roles?.includes(config.app.initialAuthorizerRole) &&  JSON.stringify(cdxUser?.facilities) === '{}')
+    {
+      //display the warning message on the home page
+      const warningMessage = 'You have no facilities for which you are responsible. Please login to CBS to review and add facilities for which you are responsible.';
+      displayAppWarning(warningMessage);
+      setWarningMessage(warningMessage);
```
yonatan-dp: warningMessage is not being used anywhere else. Why define it as a state? Perhaps use the string directly in displayWarning or only use the local variable defined right before it.

src/components/AboutHome/AboutHome.js
```diff
@@ -49,6 +52,21 @@ const AboutHome = ({ user, setCurrentLink }) => {
     );
   }, []);
 
+  //Display login warnings for the login initial authorizer user with no responsibilities.
+  useEffect(() => {
+    //get the current user
+    const cdxUser = JSON.parse(localStorage.getItem("ecmps_user"));
+
+    //check if the user has initial authorizer role with no responsibilities
+    if (cdxUser?.roles?.includes(config.app.initialAuthorizerRole) &&  JSON.stringify(cdxUser?.facilities) === '{}')
```
yonatan-dp: Doing 'JSON.stringify(cdxUser?.facilities)' to just check if facilities is empty may have overhead (among other things).  It also doesn't catch cases where facilities is null or undefined. 

Better alternative:  Object.keys(cdxUser?.facilities || {}).length === 0

For bypass users with no facilities, I think cdxUser.facilities will be null.  Can you check if you are supposed to display the warning message in such cases?

yonatan-dp (changes requested):
Please see comments...

src/components/AboutHome/AboutHome.js
EvanSun220: Hi Yonatan, the alternative way you provide is a better way to handle facilities empty/null/undefined situation. I will update them based on your notes. 
Also, Mark confirmed that we should be consistent in the behavior, so we will need to display the warning message for bypass users with no facilities.

yonatan-dp (approved):

PR #1493, opened 2024-09-26
#6369 add warning message for user with IA role and no responsibilities after logged in

### LINKED ISSUE #6369: Give user a message when they have no responsibilities

EvanSun220:
## Pull Request

- Updated AboutHome.js to display app warning message for user with initial authorizer role and no assigned responsibilities after logged in.


yonatan-dp (approved):

src/components/AboutHome/AboutHome.js
```diff
@@ -49,6 +50,20 @@ const AboutHome = ({ user, setCurrentLink }) => {
...
+  useEffect(() => {
+    //get the current user
+    const cdxUser = JSON.parse(localStorage.getItem("ecmps_user"));
+
+    //check if the user has initial authorizer role with no responsibilities
+    if (cdxUser?.roles?.includes(config.app.initialAuthorizerRole) && Object.keys(cdxUser?.facilities || {}).length === 0)
+    {
+      //display the warning message on the home page
+      const warningMessage = 'You have no facilities for which you are responsible. Please login to CBS to review and add facilities for which you are responsible.';
+      displayAppWarning(warningMessage);
```
yonatan-dp, line 68: Should the user be logged out at this point?  We have already terminated the CDX session in AUTH API and removed the session. So, we need to remove the user information in local storage (take all steps required to sign the user out). Currently, it shows that the user is logged in the UI.  See handleSignOut() in easeyAuthApi.js

yonatan-dp (changes requested):

src/components/AboutHome/AboutHome.js
EvanSun220: Hi Yonatan. From what I remembered about the discussion with you and Mark this Monday, we talked about this case. and we expect that the user will be allowed to login to ECMPS and an informational warning should be rendered on top of the homepage. (however, for ticket #6344 add certificate statement check while login, any unsigned cert statements for the user will lead to a failed login) Please let me know if my understanding needs to be corrected. Thanks.

yonatan-dp: Yep, you are correct.

yonatan-dp (dismissed):

src/components/AboutHome/AboutHome.js
```diff
@@ -49,6 +50,20 @@ const AboutHome = ({ user, setCurrentLink }) => {
     );
   }, []);
 
+  //Display login warnings for the login initial authorizer user with no responsibilities.
+  useEffect(() => {
+    //get the current user
+    const cdxUser = JSON.parse(localStorage.getItem("ecmps_user"));
+
+    //check if the user has initial authorizer role with no responsibilities
+    if (cdxUser?.roles?.includes(config.app.initialAuthorizerRole) && Object.keys(cdxUser?.facilities || {}).length === 0)
```
mark-hayward-erg: @EvanSun220 The check for no responsibilities should be applied for any users that have the Sponsor, Submitter, Preparer, or Initial Authorizer roles.  

The example user given in the ticket only had the IA role, but the check should apply to all users that can prepare or submit data.

mark-hayward-erg (changes requested):

mark-hayward-erg (approved):

PR #1504, opened 2024-10-31
fix:offline message as per the content

### LINKED ISSUE #6243: Add the ability to take ECMPS offline
Add the ability to take ECMPS offline so that people can't submit. Also, we may want a test mode where testing can be performed but outside users can't access the system.

Need to look into what can be done with the APIs.

Via the content repo, update the disabled message to state, "The ECMPS 2.0 Login has been disabled due to maintenance. Please read the notifications in home page’s "What's New" box and ECMPS Newsletter. For further assistance, please contact us.” 
- Note, there will likely be no link to our current ECMPS Beta "Contact Us" form as that will likely change, unless we can manage that link in the content repo, link TBD.

For ECMPS 1.0, we use three different status codes. 

1. DOWN - used for scheduled maintenance and migration activities. Nobody can log in.
2. ONLINE - anyone with the appropriate CBS/ECMPS credentials can log in.
3. PQATEST - used during migrations and allows **only** the esupport user to log in post migration activities and prior to bringing the application ONLINE. The main purpose is to ensure that users will be able to download the latest release update from the Oracle host database and successfully update their clients from the previous to new release version.

The current understanding is that there will not be a related esupport user in ECMPS 2.0. ECMPS Support staff will have production CDX ECMPS Admin roles which can only access the System Administrator module. 

**ECMPS 2.0 Requirements**
 - This will replace the temporary up/down solution currently in place.
 - 3 potential application statuses:
   - DOWN - used for scheduled maintenance and migration activities. Nobody can log in and all APIs will be unavailable.
   - ONLINE - anyone with the appropriate CDX credentials can log in to ECMPS 2.0.  APIs will be available.
   - TEST - used during migrations and allows only specified users to log in post migration activities and prior to bringing the application ONLINE.  API endpoints should only be accessible from the ECMPS or CAMPD UI and should not be able to be called directly.
     - List of specified users will be controlled by a new environment variable (similar to how users are specified for the CDX bypass)
       - If the application status is TEST and the CDX user id is in the list of specified test users, then the user will be allowed to log in normally
       - If the application status is TEST and the CDX user id is NOT in the list of specified test users, then the user will NOT be allowed to log in and the user will receive a message indicating the DOWN status.

ergjustin:
## Ticket
[Add the ability to take ECMPS offline](https://github.com/US-EPA-CAMD/easey-ui/issues/6243)


src/utils/api/qaCertificationsAPI.js
```diff
@@ -2,44 +2,7 @@ import axios from "axios";
...
-          "x-api-key": config.app.apiKey,
-        };
-      }
-    }
-  } catch (e) {
-    displayAppError(e);
-  }
-  return axios(options);
-};
+import { secureAxios } from "./easeyAuthApi"
```
maxdiebold-erg, line 5: This change is causing `qaCertificationAPI.test.js` to fail for me, since `refreshClientToken` is called even if `ecmpsUser` is null.

I think you could change occurrences of `expect(mock.history.get.length).toBe(1);` to `expect(mock.history.get.length).toBe(2);` or, alternatively, move the `refreshClientToken` call inside the `if (ecmpsUser) {` block.

maxdiebold-erg (changes requested):
All looks good! I did get one failing unit test, noted below.

src/utils/api/qaCertificationsAPI.js
ergjustin: I had an issue in my local environment, fixed it, and pushed it to GitHub([link](https://github.com/US-EPA-CAMD/easey-ecmps-ui/pull/1504/commits/bc0183c74d660a888f40872b7aa76a7917138d4c)). But GitHub threw an error, so I reverted the changes([link](https://github.com/US-EPA-CAMD/easey-ecmps-ui/pull/1504/commits/9092edc048c77fed3a02d5ab59320c5ddaab9875)). Now everything's working smoothly again.

mark-hayward-erg (approved):

maxdiebold-erg (approved):

PR #1509, opened 2024-11-21
6451_component_changes_not_saving_from_the_UI

### LINKED ISSUE #6451: Bug: Component changes not saving from the UI
Users are currently able to edit the following component-only fields: Manufacturer, Model or Version, Serial Number, and Hg Converter Indicator.

The changes are not saved to camdecmpswks.component after editing and saving these changes in the UI though. The changes will trigger the monitoring plan for evaluation correctly, but none of the affected columns (model_version, serial_number, manufacturer, update_date, and hg_converter_ind) are updated.

**Note:** Importing the same changes from json currently works as intended. Making/ saving the changes in the UI should do the same.

Jhernandez29:
Ticket
https://github.com/US-EPA-CAMD/easey-ui/issues/6451

Changes
Updated logic for DataTableSystems.saveComoponent by removing selectedRangeInFirst.hgConverterIndicator. The userInput.hgConverterIndicator is being over written by this value before calling PUT method.


maxdiebold-erg (approved):
Looks good!

maxdiebold-erg (changes requested):
After looking more into the `easey-monitor-plan-api` code related to this, I think the UI should be updated to update the `component` and the `monitor_system_component` record separately. Alternatively, a new endpoint could be created in the MP API that does both, but I don't think the behavior of updating the component record should be added to the existing `updateSystemCompnent` method.

If the route of updating the records separately is taken, something like the following would need to be added to `utils/api/monitoringPlansApi.js`:
```javascript
export const saveComponents = async (payload, locId, compId) => {
  const url = getApiUrl(`/locations/${locId}/components/${compId}`);

  try {
    return handleResponse(
      await secureAxios({
        method: "PUT",
        url: url,
        data: payload,
      })
    );
  } catch (error) {
    return handleImportError(error);
  }
};
```

Then, the `saveComponent` function would need to be updated to separate the payloads properly and call both `mpApi.saveSystemsComponents` and `mpApi.saveComponents`.

Jhernandez29:
Hi @maxdiebold-erg Maxwell,

Just wanted to let you know that I've already addressed the issue. Everything should be working as expected in the UI.

src/components/datatablesContainer/DataTableSystems/DataTableSystems.js
```diff
@@ -678,27 +678,40 @@ export const DataTableSystems = ({
...
+
+    if ((resp?.status >= 200 && resp?.status < 300) || (response?.status >= 200 && response?.status < 300)) {
+      setupdateComponentTable(true);
+      setUpdateRelatedTables(true);
+      setErrorMsgs([]);
+      return true;
+    } else {
+      const errorResp = Array.isArray(resp) ? resp : [resp];
+      setErrorMsgs(errorResp);
     }
```
maxdiebold-erg: If the `saveComponents` response is non-200, it will not be printed to the user here. This should be updated to check if there are error messages on that response:
```suggestion
      const newErrorMsgs = [];
      [resp, response].forEach((res) => {
        if (res?.status < 200 || res?.status >= 300) {
          newErrorMsgs.push(res);
        }
      });
      if (newErrorMsgs.length > 0) {
        setErrorMsgs(newErrorMsgs.flat());
        return false;
      }

      setupdateComponentTable(true);
      setUpdateRelatedTables(true);
      setErrorMsgs([]);
      return true;
```

maxdiebold-erg (changes requested):
Looks good, thanks for making these changes. I noticed that the error messages (if any) are not being shown for one of the responses, I've left a suggestion below.

src/components/datatablesContainer/DataTableSystems/DataTableSystems.js
Jhernandez29: Changes committed with last code change.

Jhernandez29:
Hi @maxdiebold-erg. I updated the last request to show both API error messages.

src/components/datatablesContainer/DataTableSystems/DataTableSystems.js
```diff
@@ -679,28 +678,49 @@ export const DataTableSystems = ({
...
         setupdateComponentTable(true);
         setUpdateRelatedTables(true);
         setErrorMsgs([]);
         return true;
-      } else {
-        const errorResp = Array.isArray(resp) ? resp : [resp];
-        setErrorMsgs(errorResp);
+
+      } catch (error) {
+        setErrorMsgs(JSON.stringify(error));
```
maxdiebold-erg, line 721: The original code here was correct, `setErrorMsgs` takes an array argument, not a string:
```suggestion
        setErrorMsgs([JSON.stringify(error)]);
```

maxdiebold-erg (changes requested):
The changes look good! I found one more thing that I missed initially, noted below.

src/components/datatablesContainer/DataTableSystems/DataTableSystems.js
Jhernandez29: Updated the ReactHook value as an array

maxdiebold-erg (approved):

PR #1516, opened 2024-12-24
using log level instead of console log

### LINKED ISSUE #6435: Filter out debug logs in upper environments
The **`Logger`** class from `easey-common` should be used for all logs, and the `*_ENABLE_DEBUG` environment variable should be used to conditionally limit the level of logs that are printed. If the `*_ENABLE_DEBUG` flag is False, only logs of level **error** and **log** should be output.

Additionally, review all current logs to ensure that no sensitive information is being printed (authentication tokens, for example).

All instances of printing logs with `console.<level>` should be converted to the appropriate method of the `Logger` class in API repositories that utilize Nest. In the ECMPS UI, the `loglevel` package should be used.

ergjustin:
## Ticket
[Filter out debug logs in upper environments](https://github.com/US-EPA-CAMD/easey-ui/issues/6435)

## Changes

- Using log level instead of console log

### DIFF
--- a/src/additional-functions/checkout.js
+++ b/src/additional-functions/checkout.js
@@ -1,3 +1,4 @@
+import log from "loglevel";
 import * as mpApi from "../utils/api/monitoringPlansApi";
 import { MONITORING_PLAN_STORE_NAME } from "../additional-functions/workspace-section-and-store-names";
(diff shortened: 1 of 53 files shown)


maxdiebold-erg (approved):
Looks good!

PR #1517, opened 2024-12-27
Fix/#6449 emission loading spinner missing

### LINKED ISSUE #6449: Bug: Loading spinner missing in lower grid of Emissions screen
There used to also be a loading spinner in the lower grid of the Emissions screen while data was still loading for a view.   As of now, the following is displayed to a user while the data is loading. 

Image

When user clicks the filter, "loading data" should appear until the data is ready to display.

Jhernandez29:
Ticket
https://github.com/US-EPA-CAMD/easey-ui/issues/6449

Changes
Updated EmissionsViewTable with useEffect to update the spinner when the data is loading. Added a second useEffect to update the message to display if there is no data and when the user needs to click the apply filter.

### DIFF
--- a/src/components/EmissionsTabRender/EmissionsTabRender.js
+++ b/src/components/EmissionsTabRender/EmissionsTabRender.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useCallback, useEffect, useState } from "react";
 import HeaderInfo from "../HeaderInfo/HeaderInfo";
 import "../MonitoringPlanTab/MonitoringPlanTab.scss";
 import CustomAccordion from "../CustomAccordion/CustomAccordion";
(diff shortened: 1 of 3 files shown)


src/components/EmissionsTabRender/EmissionsTabRender.js
```diff
@@ -33,10 +33,15 @@ export const EmissionsTabRender = ({
...
 
   const [viewTemplateSelect, setViewTemplateSelect] = useState(null);
+  const [filterApply, setFilterApply] = useState(false);
 
   // Determines if a user has just navigated to the page without applying any filters yet
   const isInitialLoadOfPage = currentTab?.isViewDataLoaded === undefined;
 
+  const memoizedSetFilterApply = useCallback((value) => {
+    setFilterApply(value);
+  }, [setFilterApply]);
```
maxdiebold-erg: There is no need to memoize state setters, they are stable across renders.
```suggestion
```

maxdiebold-erg (dismissed):
Works well!

I left one suggestion below: memoizing the `setFilterApply` call shouldn't be necessary, React guarantees that `setState` calls are stable across renders (but it isn't hurting anything).

src/components/EmissionsTabRender/EmissionsTabRender.js
Jhernandez29: Removed memoize state setter

maxdiebold-erg (approved):

PR #1520, opened 2025-01-08
Fix/6051 Facility Location filtering not working in Error Suppression Module

### LINKED ISSUE #6051: Facility Location filtering not working in Error Suppression Module
The Facility Name/ ID and Location Name fields do not filter correctly when searching for error suppressions. Entering a Facility Name/ ID and clicking Apply Filters does nothing, and the Location Name dropdown does not work.

## Acceptance Tests:
1. Enter a Facility Name or ID in the "Facility Name/ ID" field and click "Apply Filters".
The appropriate facility is returned.
2. The "Location Name" dropdown allows the selection of a specific Location Name. After selection, the appropriate location is returned.

Jhernandez29:
Ticket
https://github.com/US-EPA-CAMD/easey-ui/issues/6051
Changes

Updated ComboBox.jsx to clear input and selected value when clearing, updated ErrorSuppressionDataContainer.js, and updated ErrorSuppressionFilters.onFacilityChange with a new else statment if value does not equal null.

### DIFF
--- a/src/components/ComboBox/ComboBox.jsx
+++ b/src/components/ComboBox/ComboBox.jsx
@@ -68,6 +68,7 @@ const ComboBoxForwardRef = (
     ulProps,
     customFilter,
     disableFiltering = false,
+    selectedFacility
   },
   ref
 ) => {
@@ -161,6 +162,15 @@ const ComboBoxForwardRef = (
       }
     }
   }, [state.focusMode])
+
+  useEffect(() => 
+  {
+    if(typeof selectedFacility == "string" && selectedFacility == "")
+    {
+      dispatch({ type: ActionTypes.CLEAR_SELECTION })
+    }
+  }
+  ,[selectedFacility])
 
   useImperativeHandle(
     ref,
--- a/src/components/ErrorSuppression/ErrorSuppressionFilters/ErrorSuppressionFilters.js
+++ b/src/components/ErrorSuppression/ErrorSuppressionFilters/ErrorSuppressionFilters.js
@@ -205,7 +205,7 @@ export const ErrorSuppressionFilters = () => {
   };
 
   const onFacilityChange = (value) => {
-    setSelectedFacility(value === null ? null : value);
+    setSelectedFacility(value);
     if (!value || value === defaultDropdownText) {
       setSelectedLocations([]);
       setLocationData([]);
--- a/src/components/ErrorSuppression/ErrorSuppressionFilters/ErrorSuppressionFilters.js
+++ b/src/components/ErrorSuppression/ErrorSuppressionFilters/ErrorSuppressionFilters.js
@@ -236,11 +236,11 @@ export const ErrorSuppressionFilters = () => {
 
   const onCheckTypeChange = (e) => {
     let { value } = e.target;
-    value = value === "false" ? false : value;
+    value = value === "false" ? undefined : value;
 
     setSelectedCheckType(value);
-    setSelectedCheckNumber(false);
-    setSelectedCheckResult(false);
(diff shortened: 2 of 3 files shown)


src/components/ErrorSuppression/ErrorSuppressionFilters/ErrorSuppressionFilters.js
```diff
@@ -205,23 +205,33 @@ export const ErrorSuppressionFilters = () => {
   };
 
   const onFacilityChange = (value) => {
-    setSelectedFacility(value);
+    setSelectedFacility(value === null ? null : value);
```
maxdiebold-erg: This is unnecessary, it always evaluates to `value`.
```suggestion
    setSelectedFacility(value);
```

src/components/ComboBox/ComboBox.jsx
```diff
@@ -162,6 +163,15 @@ const ComboBoxForwardRef = (
...
 
+  useEffect(() => 
+  {
+    if(typeof selectedFacility == "string" && selectedFacility == "")
+    {
+      dispatch({ type: ActionTypes.CLEAR_SELECTION })
+    }
+  }
+  ,[selectedFacility])
+
```
maxdiebold-erg, line 174: What is this supposed to do? The `selectedFacility` variable is never a string, it is either a number or undefined.

maxdiebold-erg (changes requested):
The new logic for getting locations looks good, but I'm not sure about the other changes. Also, if we want to allow for fetch error suppressions without passing in the check* parameters, those parameters should be filtered out of the request object, or they should be updated to `undefined` instead of `"false"`: for the `checkNumber` in particular, passing a string causes the request to fail validation, and for `checkResult` & `checkTypeCode` the string "false" will mess up the database query.

Jhernandez29:
@maxdiebold-erg. On src\components\ErrorSuppression\ErrorSuppressionFilters\ErrorSuppressionFilters.applyFIlter method we can see  setCheckType(
      selectedCheckType !== defaultDropdownText ? selectedCheckType : null
    );
    setCheckNumber(
      selectedCheckNumber !== defaultDropdownText ? selectedCheckNumber : null
    );
    setCheckResult(
      selectedCheckResult !== defaultDropdownText ? selectedCheckResult : null
    );                                                                                                                                                                                                               This will make the check* parameters null and they are passed as params to http://localhost:8060/camd-services/admin/error-suppressions?... the check* Params will not be present. Because defaultDropdownText equals 'Select a value'. This will happen if they want to allow for fetch error suppressions without passing in the check* parameters

maxdiebold-erg:
@Jhernandez29 That does seem to be the intent, but it's not working that way because the _value_ `false` is assigned to the option that has the _label_ "-- Select a value --". So the value for those three filters can be undefined or false, but never the `defaultDropdownText` value (and so never null).

Jhernandez29:
@maxdiebold-erg  updated all check* params to undefined instead of "false"

maxdiebold-erg:
Thanks for making those changes, works well!

maxdiebold-erg (approved):

src/components/ComboBox/ComboBox.jsx
mark-hayward-erg: @maxdiebold-erg @Jhernandez29 has this conversation be resolved?  It is blocking the merging of this PR.

maxdiebold-erg: In my testing, the conditional block is never entered. I'd vote for removing this effect, but it's not hurting anything so I didn't want to block the PR over it.

PR #1522, opened 2025-01-22
Auth API changed to resolve JSON hijacking vulnerability by returning an object

### LINKED ISSUE #6519: JSON Hijacking Vulnerable Endpoints - Easey-Auth-API
# JSON Hijacking Vulnerability Analysis & Solution

## Overview
JSON hijacking vulnerability has been identified in several endpoints that return arrays directly via GET requests in the easey-auth-api. This document outlines the vulnerable endpoints and provides a solution to mitigate this security risk.

**Note**:  Please see #6518 for corresponding changes to performance test scripts.

---

### 1. Certification Related 
```
- api.epa.gov/easey/staging/auth-mgmt/statements
```
---

### 1. Use Response Interface

Create a new file `src/interfaces/array-response.interface.ts`:
**Or better, create this interface in easey-common for sharing** 

```typescript
export interface ArrayResponse<T> {
  data: T[];
}
```

---

### 2. Example usage 

Here's how to modify the `CertificationsController` to wrap the array response:

```typescript
@Get('/statements')
@ApiOkResponse({
  description: 'Returns a list of certification statements',
  type: ArrayResponse<CertificationStatementDTO>
})
@ApiQuery({
  style: 'pipeDelimited',
  name: 'monitorPlanIds',
  required: true,
  explode: false,
})
async statements(
  @Query() dto: CertificationParamDTO,
): Promise<ArrayResponse<CertificationStatementDTO>> {
  const statements = await this.service.getStatements(dto.monitorPlanIds);
  return {
    data: statements
  };
}
```

**Note**: Make sure swagger documentation is updated along with code changes. 

---

## Frontend Changes Required

Since the backend is now returning an array wrapped in an object, the frontend or any consumer of this API endpoints needs to be updated to handle this new response type. Assuming the frontend is using the common handleResponse function (easey-camd-ui for instance), this common function can be updated as follows to handle both `array` and `non-array` response types.

#### Before
```javascript
return axios.get(url).then(handleResponse).catch(handleError);

export async function handleResponse(response) {
  if (response.status === 200 || response.status === 201) {
    return response;  // response.data was an array
  } else {
    throw new Error("failed");
  }
}
```

#### After
```javascript
return axios.get(url).then(handleResponse).catch(handleError);

export async function handleResponse(response) {
  if (response.status === 200 || response.status === 201) {
    // Check if response is a wrapped array
    if (response.data && Array.isArray(response.data.data)) {
      return response.data.data;  // unwrap array from response object
    }
    // Return original response data for non-array responses
    return response.data;
  } else {
    throw new Error("failed");
  }
}
```

## Testing

Update existing tests to expect the new response structure. For example:

```javascript
it('should return a wrapped array response', async () => {
  const expectedResult = {
    data: [] as CertificationStatementDTO[]
  };
  
  jest.spyOn(service, 'getStatements')
    .mockResolvedValue(expectedResult.data);
    
  expect(await controller.statements(mockCertificationParamDTO))
    .toEqual(expectedResult);
});
```

**Note**: The above code is provided as sample and has not been tested. Please adjust as needed and test thoroughly.

yonatan-dp:
Auth API changed to resolve JSON hijacking vulnerability by returning an ArrayResponse object instead of returning an array directly from an API. ECMPS UI updated to handle that.

### DIFF
--- a/src/components/SubmissionModal/SubmissionModal.js
+++ b/src/components/SubmissionModal/SubmissionModal.js
@@ -54,7 +54,7 @@ export const SubmissionModal = ({
       setLoading(true);  // Set loading to true before fetching data
       try {
         const result = await getCredentials(monitorPlanIds);
-        const list = result.data.data.map((el) => ({
+        const list = result.data.items.map((el) => ({
           title: "Certification Statement",
           content: el.statementText,
           expanded: false,
--- a/src/components/SubmissionModal/SubmissionModal.js
+++ b/src/components/SubmissionModal/SubmissionModal.js
@@ -54,7 +54,7 @@ export const SubmissionModal = ({
       setLoading(true);  // Set loading to true before fetching data
       try {
         const result = await getCredentials(monitorPlanIds);
-        const list = result.data.map((el) => ({
+        const list = result.data.data.map((el) => ({
           title: "Certification Statement",
           content: el.statementText,
           expanded: false,
--- a/src/utils/api/easeyAuthApi.test.js
+++ b/src/utils/api/easeyAuthApi.test.js
@@ -254,9 +254,9 @@ describe("Easey Auth API", () => {
         `${config.services.authApi.uri
         }/certifications/statements?monitorPlanIds=${monitorPlans.join("|")}`
       )
-      .reply(200, { data: {} });
+      .reply(200, { items: {} });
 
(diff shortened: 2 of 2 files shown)


ergjustin (approved):
Looks good to me

PR #1531, opened 2025-02-10
vite implement

### LINKED ISSUE #6545: Replace use of Create React App with Vite
ecmps-ui and campd-ui both currently use Create React App, which has a vulnerable dependency on react-scripts.  

This dependency is outdated and appears to no longer be maintained.  A number of our open Dependabot alerts include react-scripts as a dependency that is blocking resolution.

A number of other ERG-maintained react apps have switched to Vite, and this appears to be a widely accepted build tool.

### Requirements:  
1. Replace use of Create React App with Vite
2. Update testcase with jest.
3. Update uswds css with uswds-compile

ergjustin:
## Ticket
[Replace use of Create React App with Vite](https://github.com/US-EPA-CAMD/easey-ui/issues/6545)

## TODO

- [ ] Need to update environment variables


maxdiebold-erg (changes requested):
This runs well locally! I know you're still working on this (especially the styles), but I'll add a couple points:
  - `craco.config.js` can be removed, since that package is no longer installed.
  - I'm seeing a lot of console warnings related to Babel when installing the new dependencies. They don't lead to any issues, and it looks like Babel is only used for the Jest unit tests. We may want to consider moving to Vitest eventually, but that's a heavy lift and may not be any benefit. In the meantime, the majority of these warnings can be resolved by explicitly including `@babel/core` in the `devDependencies`.

package.json
```diff
@@ -12,11 +12,12 @@
...
   "scripts": {
-    "start": "craco start",
-    "build": "craco build",
-    "test": "craco test --coverage --testResultsProcessor=jest-sonar-reporter",
-    "test:cov": "craco test --coverage",
+    "start": "vite",
+    "build": "vite build",
+    "test": "jest --coverage --testResultsProcessor=jest-sonar-reporter",
+    "test:cov": "jest --coverage",
     "up": "yarn && yarn start"
```
maxdiebold-erg: Could you add a "prepare" script here? This is a special lifecycle script that will run when `yarn install` is run for the first time, ensuring the USWDS files are created:
```suggestion
    "prepare": "gulp init",
    "start": "vite",
    "build": "vite build",
    "test": "jest --coverage --testResultsProcessor=jest-sonar-reporter",
    "test:cov": "jest --coverage",
    "up": "yarn && yarn start"
```

maxdiebold-erg (changes requested):
Looks great! I think a line should be added to the README in the installation section to inform developers how to generate the USWDS assets. Also, in the same vein, I left a suggestion below for a script/hook to generate them on the first install.

maxdiebold-erg (dismissed):
Works great, thanks for adding those changes!

mark-hayward-erg:
@ergjustin Please resolve the conflicts that are preventing the merging of this PR.

mark-hayward-erg (approved):

PR #1532, opened 2025-02-10
feature/#6072 Add Download CSV Button for All System Admin Pages

### LINKED ISSUE #6072: Add export to csv/excel feature for System Administration
Add the ability to export to csv/excel for the various screens and reports in the System Administration tool. Excel is preferred, but csv is ok if that is easier/quicker.

## Acceptance Test
It should be possible to export all System Admin Tool screens/reports to Excel/csv.

EvanSun220:
### **Related Ticket:**
- ticket [#6072](https://github.com/orgs/US-EPA-CAMD/projects/2/views/51?filterQuery=6072&sliceBy%5Bvalue%5D=EvanSun220&pane=issue&itemId=53268922&issue=US-EPA-CAMD%7Ceasey-ui%7C6072)
### **Updates:**
- Added "Download to CSV" button and related logic for all three screens of System Admin tool

### DIFF
--- a/src/components/AdminMaintenance/EmSubmissionData/EmSubmissionData.js
+++ b/src/components/AdminMaintenance/EmSubmissionData/EmSubmissionData.js
@@ -2,7 +2,7 @@ import React, { useCallback, useState, useEffect } from "react";
 import DataTable from "react-data-table-component";
 import { Preloader } from "@us-epa-camd/easey-design-system";
 import { Button, Checkbox } from "@trussworks/react-uswds";
-import { ArrowDownwardSharp } from "@material-ui/icons";
+import { ArrowDownwardSharp, FilterDrama } from "@material-ui/icons";
 import { submissionAccessTitle } from "../../../utils/constants/moduleTitles";
 import { EmSubmissionModal } from "../EmSubmissionPopOut/EmSubmissionPopout";
 import "./EmSubmissionData.scss";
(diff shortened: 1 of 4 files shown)


src/components/AdminMaintenance/EmSubmissionData/EmSubmissionData.js
```diff
@@ -64,6 +64,57 @@ export const EmSubmissionData = ({
...
+    const a = document.createElement('a');
+    a.href = url;
+    a.download = fileName;
+    document.body.appendChild(a);
+    a.click();
+  
+    // Cleanup
+    document.body.removeChild(a);
+    URL.revokeObjectURL(url);
+  };
```
yonatan-dp, line 94: @EvanSun220 , this downloadFilteredDataIntoCSV () is duplicated across all three changes and it looks like they all do the same thing but only differ in the data they process, the column mappings, and file name prefix.  Can we not externalize this function in a utility and just invoke it?  E.g. exportToCSV(data, columnMapping, fileNamePrefix)) ....

src/components/AdminMaintenance/QAMaintenance/QAMaintenanceData.js
```diff
@@ -50,6 +50,88 @@ const QAMaintenanceData = ({
...
+      };
+    }
+  
+    const headers = Object.keys(columnMapping);
+    const csvHeaders = headers.map(key => columnMapping[key]);
+  
+    // Convert data to CSV format
+    const csvRows = filteredData.map(row =>
+      headers.map(header => (row[header] !== null && row[header] !== undefined ? `"${row[header]}"` : '""')).join(',')
+    );
```
yonatan-dp: I was wondering if it wouldn't be safe and easy to use a standard library to convert the data to csv. Standard libraries provide proper error checking, etc.  For instance `json2csv`

import { Parser } from 'json2csv';

...

// Convert the mapping to fields for json2csv
  const fields = Object.keys(columnMapping).map((key) => ({
    label: columnMapping[key],
    value: key,
  }));

  const opts = { fields };
 
try {
    const parser = new Parser(opts);
    const csv = parser.parse(filteredData);

...
}

Please look in to this and see if it is a better solution 👍

yonatan-dp (changes requested):

yonatan-dp (approved):

PR #1534, opened 2025-02-13
Fix/#6449 loading spinner message

### LINKED ISSUE #6449: Bug: Loading spinner missing in lower grid of Emissions screen

Jhernandez29:
ticket
https://github.com/US-EPA-CAMD/easey-ui/issues/6449#issuecomment-2656905205

changes 
Updated message when no data is present. Instead of adding no data record and rendering another message after 1 second. The message is static.

### DIFF
--- a/src/components/EmissionsViewTable/EmissionsViewTable.js
+++ b/src/components/EmissionsViewTable/EmissionsViewTable.js
@@ -38,7 +38,7 @@ export const EmissionsViewTable = ({ monitorPlanId, filterApply, setFilterApply 
         else
         {
             setViewData(reduxCurrentTab?.viewData || []);
-            setMessage('Select a Reporting Period, Location, and Template and Apply Filter to view the data');
+            setMessage('Select a Reporting Period, Location, and Template, then click the Apply Filter button to view the data.');
         }
     }, [reduxCurrentTab.viewColumns, reduxCurrentTab.viewData, filterApply]);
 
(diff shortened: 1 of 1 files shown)


src/components/EmissionsViewTable/EmissionsViewTable.js
```diff
@@ -38,7 +38,7 @@ export const EmissionsViewTable = ({ monitorPlanId, filterApply, setFilterApply
         else
         {
             setViewData(reduxCurrentTab?.viewData || []);
-            setMessage('Select a Reporting Period, Location, and Template and Apply Filter to view the data');
+            setMessage('Select a Reporting Period, Location, and Template, then click the Apply Filter button to view the data.');
```
maxdiebold-erg, line 41: For consistency, I'd say either add the trailing period everywhere or leave it off. The initial message does **not** have a trailing period.

src/components/EmissionsViewTable/EmissionsViewTable.js
```diff
@@ -47,10 +47,9 @@ export const EmissionsViewTable = ({ monitorPlanId, filterApply, setFilterApply
         {
             if ( viewData?.length === 0) {
                 const timeOutApply = setTimeout(() => {
-                    setMessage('Select a Reporting Period, Location, and Template and Apply Filter to view the data');
                     setPendingMessage(false)
                 }, 1000);
-                setMessage('There are no records to display');
+                setMessage('There are no records to display.\nSelect another Reporting Period, Location, and Template, then click the Apply Filter button to view the data.');
```
maxdiebold-erg: You could just set the HTML here and avoid the changes below:
```suggestion
        setMessage(
          <div>
            <p>There are no records to display</p>
            <p>
              Select another Reporting Period, Location, and Template, then
              click the Apply Filter button to view the data
            </p>
          </div>
        );
```
This also avoids the extra empty `div` that is being added in the `customNoDataMessage`.

maxdiebold-erg (dismissed):
Looks good, thanks for updating that text!

I left a couple suggestions, but they're nitpicks so I'll leave the call up to you.

maxdiebold-erg (approved):

PR #1539, opened 2025-02-19
Fix JSON hijacking vulnerability defect

yonatan-dp:
Fix JSON hijacking vulnerability defect


ergjustin (approved):
Looks good to me
Able to checkout and check in in public view

PR #1543, opened 2025-02-21
fix:selectConfigurations button color and header background color

### LINKED ISSUE #6545: Replace use of Create React App with Vite

ergjustin:
## Ticket
[Replace use of Create React App with Vite](https://github.com/US-EPA-CAMD/easey-ui/issues/6545)

## Changes

- SelectConfigurations button color 
- Header background color

### DIFF
--- a/src/components/MonitoringPlanTab/MonitoringPlanTab.js
+++ b/src/components/MonitoringPlanTab/MonitoringPlanTab.js
@@ -59,7 +59,7 @@ export const MonitoringPlanTab = ({
   }, [locationSelect]);
 
   return (
-    <div>
+    <div id="monitoring-plan-tab-container">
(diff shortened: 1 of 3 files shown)


maxdiebold-erg (approved):
Good catch!

PR #1545, opened 2025-02-21
Configuration Management: Evaluation & Submission Updates #6593

### LINKED ISSUE #6593: Configuration Management: Evaluation & Submission Updates
If a user performs any of the actions below via Configuration Management, **all** active monitoring plans at the facility are marked for needs evaluation and resubmission. 

- Create a new monitoring plan for a new unit. 
- Add new stack/pipe and associated USC records. 
- End stack/pipe and associated USC records. 

The only monitoring plans that should be required for evaluation and resubmission are those affected by the configuration changes. In ECMPS 1.0, this would also include other, already inactive monitoring plans that a unit was once associated with.

This is also not an issue if these same actions are done via json import.

Jhernandez29:
Ticket:
https://github.com/US-EPA-CAMD/easey-ui/issues/6593

Changes:
I added a comparison function where the stackPipe and unitStackConfigs are filtered using OriginalRecord. The units are filtered by using the isToggle value when set to true a new monitoring plan wants to be added, and also another filter we have to do is comparing filteredUnitStackConfigs with units because unitStacksConfigs were added to a new stack pipe.
Those monitoring plans, the single unit ones end and then were added to a new monitoring plan with the stack pipe.

### DIFF
--- a/src/components/ConfigurationManagement/ConfigurationManagement.js
+++ b/src/components/ConfigurationManagement/ConfigurationManagement.js
@@ -886,6 +886,28 @@ export const ConfigurationManagement = ({
     });
   };
 
+  const filterFormState = () => {
+    const filteredStackPipes = formState.stackPipes.filter(
+      stackPipe => stackPipe?.retireDate !== stackPipe?.originalRecord?.retireDate
+      || stackPipe?.originalRecord == null
+    );
+    const filteredUnitStackConfigs = formState.unitStackConfigs.filter(
+      unitStackConfig => unitStackConfig?.endDate !== unitStackConfig?.originalRecord?.endDate
+      || unitStackConfig?.originalRecord == null
+    );
+    const filteredUnits = formState.units.filter(unit => 
+      unit.isToggled === true || 
+      filteredUnitStackConfigs.some(config => config.unitId === unit.unitId)
+    );
+    
+    return {
+      units: filteredUnits,
+      stackPipes: filteredStackPipes,
+      unitStackConfigs: filteredUnitStackConfigs
+    };
+  };
+  
+
   const initializeEditableFormState = (data, type) => {
     formDispatch({
       type,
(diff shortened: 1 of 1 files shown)


src/components/ConfigurationManagement/ConfigurationManagement.js
```diff
@@ -886,6 +886,28 @@ export const ConfigurationManagement = ({
...
+      unit.isToggled === true || 
+      filteredUnitStackConfigs.some(config => config.unitId === unit.unitId)
+    );
+    
+    return {
+      units: filteredUnits,
+      stackPipes: filteredStackPipes,
+      unitStackConfigs: filteredUnitStackConfigs
+    };
+  };
```
maxdiebold-erg: Can just return the stack/pipes and USCs here:
```suggestion
    
    return {
      stackPipes: filteredStackPipes,
      unitStackConfigs: filteredUnitStackConfigs
    };
  };
```

maxdiebold-erg (changes requested):
Looks great! The stack/pipes and USCs are being filtered as they should be. I would only add that I think the units should be filtered out completely: the **only** thing that can be done to a unit is to associate it with a new monitoring plan if it hasn't been associated with one previously, and that is handled in the `sendSingleUnitPayloads` function.

As it stands, I don't think including the units could affect anything that wouldn't already be affected, but in case there is some edge case not being considered, it's probably best to just pull them out.

In addition to the suggestion below, lines 945 - 954 can be dropped:
```diff
-      .concat(
-        formState.units.map((u) => ({
-          unitId: u.unitId,
-          stackPipeId: null,
-          activeDate: null,
-          retireDate: null,
-          nonLoadBasedIndicator: u.nonLoadBasedIndicator,
-          ...unusedMonitoringLocationDataFields(),
-        }))
-      ),
```

src/components/ConfigurationManagement/ConfigurationManagement.js
Jhernandez29: Hi @maxdiebold-erg .     const filteredUnits = formState.units.filter(unit => 
      unit.isToggled === true || 
      filteredUnitStackConfigs.some(config => config.unitId === unit.unitId)
    );  I can delete the "unit.isToggled === true || ". I get an error when I don't include the ending single unit ones. I strongly think we should leave "filteredUnitStackConfigs.some(config => config.unitId === unit.unitId)" This creates a new monitoring plan for the unitStackConfigs that are ending.

maxdiebold-erg: Good point, I forgot about that check! Then yes, I think removing `unit.isToggled === true ||` is sufficient for the units. However, I now think that logic should be added to the `filteredStackPipes`, too, because there may be cases where a USC is ended but a stack/pipe is not retired, in which case we'll still need to include the stack/pipe in the payload.
```javascript
    const filteredStackPipes = formState.stackPipes.filter(
      (stackPipe) =>
        filteredUnitStackConfigs.some(
          (config) => config.stackPipeId === stackPipe.stackPipeId
        ) ||
        stackPipe?.retireDate !== stackPipe?.originalRecord?.retireDate ||
        stackPipe?.originalRecord == null
    );
```

maxdiebold-erg (approved):

PR #1547, opened 2025-02-25
Fix where TEE data is not automatically selected when EM data is selected

### LINKED ISSUE #6574: Submit: TEE data not automatically selected when EM data is selected
In the Submit screen, selecting EM data only will also automatically select any available MP, QAT, and QCE data. However, available TEE data is not automatically selected.

Update the logic to also include TEE data for this scenario.

yonatan-dp:
Fixed a defect where, in the Submit screen, selecting EM data only will also automatically select any available MP, QAT, and QCE data. However, available TEE data is not automatically selected.
Updated the logic to also include TEE data for this scenario.


ergjustin (approved):
Looks good to me

PR #1554, opened 2025-03-07
fix/#6079 Sorting and Filtering Issues in EM Submission Access

### LINKED ISSUE #6079: Sorting and filtering issues in Emission Submission Access
There are sorting and filtering issues in System Administration - Emission Submission Access.

The Status filter list needs to be updated. It is limited to Open, Closed and Pending Approval. It should reflect:

- **Cancelled**
  - Submission Availability Code equals 'DELETE'.
- **Closed**
  - Submission Availability Code does not equal  'GRANTED', 'REQUIRE' nor 'DELETE', and
  - EM Status Code does not equal 'PENDING', and
  - If Submission Availability Code is NULL, End Date is before Today's date.
- **No Window**
  - No EM Submission Window rows exist for the MP and reporting period combination.
  - For ECMPS 1.0, only seen in testing environments because the Window Job would normally create the windows.
- **Not Yet Open**
  - Submission Availability Code is NULL and Access End Date is on or after Today's date.
- **Open**
  - Submission Availability Code equals 'GRANTED' or 'REQUIRE'.
- **Pending Approval**
  - EM Status Code equals 'PENDING'.

Users should be able to display data by: 

~~- Facility only~~
~~- Reporting Period only~~
- Facility + Reporting Period
- Facility + Reporting Period + Status
~~- Reporting Period + Status~~

Configuration can be removed from the top grid and Facility Name and Configuration should be included as columns in the lower grid.

Filters should be added to the lower grid for:

- Facility
- Configuration
- Reporting Period
- Status
 
Column sorting also needs to be corrected. Sorting on a column with multiple rows of data and then selecting the View button causes an issue. The data in the popup is not referencing the selected row.

Users need to be able to apply the filters to the lower grid and perform secondary column sorts.

EvanSun220:
### **Related Ticket:**

- ticket [#6079](https://github.com/US-EPA-CAMD/easey-ui/issues/6079)

### **Updates:**

- Updated the logic to enable displaying all the active and inactive configurations in the new configuration filter.

### DIFF
--- a/src/components/AdminMaintenance/EmSubmissionData/EmSubmissionData.js
+++ b/src/components/AdminMaintenance/EmSubmissionData/EmSubmissionData.js
@@ -171,11 +171,9 @@ export const EmSubmissionData = ({
 
     const configNamesToMonPlan = [];
     for (const cd of configurationData) {
-      if (cd.active) {
-        const key = `${cd.facilityName} - ${cd.name}`;
-        if (!configNamesToMonPlan[key]) {
-          configNamesToMonPlan[key] = cd.id;
-        }
+      const key = `${cd.facilityName} - ${cd.name}`;
+      if (!configNamesToMonPlan[key]) {
+        configNamesToMonPlan[key] = cd.id;
       }
     }
 


Jhernandez29 (approved):
According to Scott we should allow all locations despites the active value.

PR #1555, opened 2025-03-10
Fix/#6539-Crit2-Submission-Not-Included-Errors

spetros-do:
### Changes Made: 
Updated the filtering logic in EvaluateAndSubmit.js file to only filter out CRIT1 errors only when 
"Exclude Fiels with Critical Errors" is selected.

### DIFF
--- a/src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
+++ b/src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
@@ -441,7 +441,7 @@ export const EvaluateAndSubmit = ({
           (f) => f.monPlanId === monPlanId && f.isSelected
         );
 
-        if (plan.evalStatusCode === "ERR") {
+        if (plan.evalStatusCode === "ERR" && plan.severityCode === "CRIT1") {
           payload.hasCritErrors = true;
         }
 
@@ -454,7 +454,7 @@ export const EvaluateAndSubmit = ({
       newItem.testSumIds = qaTestSumRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
             payload.hasCritErrors = true;
           }
 
@@ -463,7 +463,7 @@ export const EvaluateAndSubmit = ({
       newItem.qceIds = qaCertEventRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
             payload.hasCritErrors = true;
           }
 
@@ -472,7 +472,7 @@ export const EvaluateAndSubmit = ({
       newItem.teeIds = qaTeeRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
             payload.hasCritErrors = true;
           }
 
@@ -483,7 +483,7 @@ export const EvaluateAndSubmit = ({
       newItem.emissionsReportingPeriods = emissionsRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
             payload.hasCritErrors = true;
           }
           return m.periodAbbreviation;
@@ -527,7 +527,7 @@ export const EvaluateAndSubmit = ({
         modalType.current = "error";
         modalHeading.current = "Error";
         modalMessage.current =
-        e.response?.data?.message ?? "Error: An error occurred during the file queueing process. Please try again. If you continue to encounter an error, please contact ECMPS Support.";
+          "Error: An error occurred during the file queueing process. Please try again. If you continue to encounter an error, please contact ECMPS Support.";
         setShowSuccessModal(true);
       } else {
         handleError(e);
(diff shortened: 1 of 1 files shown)


spetros-do:
### Related To: 
(https://github.com/US-EPA-CAMD/easey-camd-services/pull/345)

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
```diff
@@ -527,7 +527,7 @@ export const EvaluateAndSubmit = ({
         modalType.current = "error";
         modalHeading.current = "Error";
         modalMessage.current =
-        e.response?.data?.message ?? "Error: An error occurred during the file queueing process. Please try again. If you continue to encounter an error, please contact ECMPS Support.";
+          "Error: An error occurred during the file queueing process. Please try again. If you continue to encounter an error, please contact ECMPS Support.";
         setShowSuccessModal(true);
```
yonatan-dp, line 531: It looks like this change is no longer looking at e.response?.data?.message.  Is this intentional?

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
```diff
@@ -738,7 +738,7 @@ export const EvaluateAndSubmit = ({
         }
       }
 
-      const dataListLength = isForceReEvaluation ? dataList.length : 4;
+      const dataListLength = isForceReEvaluation ? dataList.length : 3;
```
yonatan-dp, line 741: This also looks like change not directly tied to the ticket.  Is this intentional?

yonatan-dp (changes requested):

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
spetros-do: Yes, this change is intentional. The update removes the dependency on exposureData?.data?.message to avoid potential issues where the message may not always be populated as expected. Instead, the new logic ensures that submission errors are properly evaluated without relying on that specific data structure. And if there’s a use case where this reference is still required, I can make the adjustment.

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
spetros-do: This change was intentional to better reflect the actual size of the dataset being processed. However, I can revert it to keep the scope of the change minimal.

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
yonatan-dp: @spetros-do I think we should leave this line as is. Your response did not address why we are removing the dependency on exposureData?.data?.message.

spetros-do: @yonatan-dp I am not sure I understand what you mean, "Your response did not address why we are removing the dependency on exposureData?.data?.message." I am providing you my specific explanation why I suggest the changes made. 
Again, the purpose of removing exposureData?.data?.message is to prevent reliance on a potentially inconsistent data structure. For example, in some cases, message may not always be present, which could lead to missing error handling. 

The new logic ensures that the error message is always properly set, regardless of whether message exists in exposureData?.data?.message. If this specific reference is still required for certain scenarios, advise and I can revert it accordingly.

yonatan-dp: @spetros-do 
So, the old logic sets the modal message to `e.response?.data?.message`.  If there is no data in the exception message, then it displays the standard message of ""Error: An error occurred during the file queueing process. Please try again. If you continue to encounter an error, please contact ECMPS Support."

The new change, by removing `e.response?.data?.message` will always display the standard message, regardless of what the exception message is.  Given that this specific change is not related to the ticket description, I am asking if this is intentional (to hide the exception message). I am only being careful that we won't brake something that was put in intentionally. Also,  `e.response?.data?.message` already handles if messages are not present.

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
```diff
@@ -454,7 +454,7 @@ export const EvaluateAndSubmit = ({
       newItem.testSumIds = qaTestSumRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
```
djw4erg, line 457: Severity Codes of both FATAL and CRIT1 should be treated as critical errors.

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
```diff
@@ -463,7 +463,7 @@ export const EvaluateAndSubmit = ({
       newItem.qceIds = qaCertEventRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
```
djw4erg, line 466: Severity Codes of both FATAL and CRIT1 should be treated as critical errors.

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
```diff
@@ -472,7 +472,7 @@ export const EvaluateAndSubmit = ({
       newItem.teeIds = qaTeeRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
```
djw4erg, line 475: Severity Codes of both FATAL and CRIT1 should be treated as critical errors.

src/components/EvaluateAndSubmit/EvaluateAndSubmit.js
```diff
@@ -483,7 +483,7 @@ export const EvaluateAndSubmit = ({
       newItem.emissionsReportingPeriods = emissionsRef.current
         .filter((f) => f.monPlanId === monPlanId && f.isSelected)
         .map((m) => {
-          if (m.evalStatusCode === "ERR") {
+          if (m.evalStatusCode === "ERR" && m.severityCode === "CRIT1") {
```
djw4erg, line 486: Severity Codes of both FATAL and CRIT1 should be treated as critical errors.

djw4erg (changes requested):

spetros-do:
Closing this PR -- please see PR#355 for reference. 
https://github.com/US-EPA-CAMD/easey-camd-services/pull/355

PR #1563, opened 2025-04-02
To allow MP to begin dates up to 90 days in future. (fix/#6534-allow-future-begin-dates)

spetros-do:
### Changes Made:
- Updated backend MAXIMUM_FUTURE_DATE constant from 30 to 90 days
- Modified frontend date picker to allow future date selection up to 90 days for:
    - Defaults
    - Formulas
    - Loads
    - Location Attributes
    - Methods
    - Qualifications
    - Rectangular Duct WAFs
    - Spans
    - Systems
    - Unit Information

### Testing:
- Verify users can select and save begin dates up to 90 days in the future on all specified screens.

```

### CODE REVIEW

**Comment by spetros-do** [comment]
https://github.com/US-EPA-CAMD/easey-ui/issues/6534

File: src/components/ModalDetails/ModalDetails.js [hunk 1]
```diff
@@ -376,11 +376,20 @@ const ModalDetails = ({
...
             maxDate={
-              title !== "Protocol Gas"
-                ? `${d.getFullYear()}-${d.getMonth() + 1}-${d.getDate()}`
-                : null
-            }
-          />
+                  // Allow future dates up to 90 days for specific components
+                  ["Defaults", "Formulas", "Loads", "Location Attributes", "Methods",
+                    "Qualifications", "Rectangular Duct WAFs", "Spans", "Systems",
+                    "Unit Information"].includes(title)
```
ergjustin: Could you please remove the extra 's' from the following title list?

- Defaults → Default
- Formulas → Formula
- Loads → Load
- Location Attributes → Location Attribute
- Methods → Method
- Qualifications → Qualification
- Rectangular Duct WAFs → Rectangular Duct WAF
- Spans → Span
- Systems → System
- Create Systems → Create System

Also, I’m not sure if all Unit information needs to be included—could you kindly confirm? If needed, please ensure the array list includes: 

- Unit Fuel
-  Unit Control
- Unit Capacity

ergjustin (changes requested):

src/components/ModalDetails/ModalDetails.js
spetros-do: - Thanks for catching the array list count and plural format. 
- The update is not just for Protocol Gas; in the original code, it had a special case for "Protocol Gas", where maxDate was set to null -- meaning no date restriction; while all other components had a maxDate set to the current date. 
- Based on the requirement described on the change made maintains the special case for "Protocol Gas"; but adds a new condition for the components that should allow users to select dates up to 90 days in the future.

spetros-do:
Replaced by PR: https://github.com/US-EPA-CAMD/easey-ecmps-ui/pull/1571

PR #1590, opened 2025-05-16
fix/#6687 Missing Configuration Name in MP Exports

### LINKED ISSUE #6687: Missing configuration name in MP exports
When exporting a Monitoring Plan, the JSON filename is showing "undefined" in place of the configuration name, e.g. "MP Export - Greene County, undefined (5-1-2025).json".

This is due to UI expecting the "name" field in the export payload from the Monitor Plan API, but the API process removes it along with other non-reported values. If this is correct, then the configuration name should be instead drawn from UI state.

EvanSun220:
### **Related Ticket:**

- ticket [#6687](https://github.com/US-EPA-CAMD/easey-ui/issues/6687)

### **Updates:**

- Updated the signature of monitorPlansApi.exportMonitorPlanDownload method to include the config name which will be used as a part of exported file name.
- Updated the impacted logic in HeaderInfo.js and ExportTab.js.


maxdiebold-erg (approved):
Looks good!

PR #1591, opened 2025-05-16
remove Calculated Rows from Quarterly Summary View in Public View

### LINKED ISSUE #6714: Remove Calculated Rows from Quarterly Summary View in Public View
Calculated columns were removed from the public view UI in #6609. "Report Errors" was the only column removed from the Quarterly Summary View. This screen also includes rows for "Quarterly Calculated", "Year to Date Calculated", and "Ozone Season Calculated" that we may also want to consider removing in the future.

Image

ergjustin:
## Ticket
[Remove Calculated Rows from Quarterly Summary View in Public View](https://github.com/US-EPA-CAMD/easey-ui/issues/6714)

## Changes
Emission table row update with filtering process


Jhernandez29:
@ergjustin  This ticket is to remove duplicates or remove columns depending on the rules? 

if you want to remove the duplicated columns depending on rowname you could use this :
const seen = new Set();
const uniqueData = data.filter(row => {
  if (seen.has(row.rowName)) {
    return false;
  }
  seen.add(row.rowName);
  return true;
});

ergjustin:
> @ergjustin This ticket is to remove duplicates or remove columns depending on the rules?
> 
> if you want to remove the duplicated columns depending on rowname you could use this : const seen = new Set(); const uniqueData = data.filter(row => { if (seen.has(row.rowName)) { return false; } seen.add(row.rowName); return true; });

@Jhernandez29 
The ticket specified that rows containing the '**Calculated**' column should be removed in the global view

Jhernandez29 (approved):
The ROW_FILTER_CONFIGS pattern introduces a structured and maintainable approach to row filtering. This abstraction is likely to be very useful as the application evolves.

PR #1595, opened 2025-05-23
Updated code to allow UI to begin dates up to 90 days in the future.

spetros-do:
### Changes Made:
Updated UI for MAXIMUM_FUTURE_DATE constant from 30 to 90 days by adding a centralized future date logic using utility functions [getMaxFutureDate() and formatDateToInput()] from [futureDateControls.js.] to allow future date selection up to 90 days for:
- Default
- Formula
- Load
- Location Attribute
- Method
- Qualification
- Rectangular Duct WAF
- Span
- System
- Unit Information's: 
  - Unit Fuel
  - Unit Control
  - Unit Capacity

Testing:
Verified users can select and save begin dates up to 90 days in the future on all specified screens.

Ref:
PR: #1563 
Issue: US-EPA-CAMD/easey-ui#6534


src/components/ModalDetails/ModalDetails.js
```diff
@@ -16,6 +16,12 @@ import { ArrowBackSharp } from "@material-ui/icons";
...
 
+import {
+  FUTURE_DATE_ALLOWED_COMPONENTS,
+  getMaxFutureDate,
+  formatDateToInput
+} from "../../utils/constants/futureDateControls";
+
 // value in data => [0] api label, [1] our UI label, [2] value,[3], required or not for editing, [4] control form type
 const ModalDetails = ({
   modalData,
```
maxdiebold-erg, line 27: The would be added here:
```suggestion
const ModalDetails = ({
  allowFutureDates,
  modalData,
```

src/components/ModalDetails/ModalDetails.js
```diff
@@ -376,20 +381,13 @@ const ModalDetails = ({
...
-                    const futureDate = new Date(d.getTime() + (90 * 24 * 60 * 60 * 1000));
-                    return `${futureDate.getFullYear()}-${futureDate.getMonth() + 1}-${futureDate.getDate()}`;
-                  })()
-
-              : title !== "Protocol Gas"
-                ? `${d.getFullYear()}-${d.getMonth() + 1}-${d.getDate()}`
-                : null
-            }
-          />
+              FUTURE_DATE_ALLOWED_COMPONENTS.includes(title)
```
maxdiebold-erg, line 384: The new prop would be used here:
```suggestion
              allowFutureDates
```

maxdiebold-erg (changes requested):
The refactoring here is nice! However, the `FUTURE_DATE_ALLOWED_COMPONENTS` array does not include the missing components noted in https://github.com/US-EPA-CAMD/easey-ui/issues/6534#issuecomment-2855889621, and it's actually changed the component titles from the previous changes (by making many of them plural) such that only three allow selecting a future date.

However, take a look at how **`DataTableSystems`** uses the **`ModalDetails`** component:
```javascript
<ModalDetails                                 
  modalData={selected}                        
  data={selectedModalData}                    
  cols={2}                                    
  title={`System: ${selected[0]["value"]}`}   
  viewOnly={!(user && checkout)}              
  setDisableExitBtnStatus={setDisableExitBtn} 
/>                                            
```
The `title` prop here is dynamically generated, so there's no way to add it to the literal `FUTURE_DATE_ALLOWED_COMPONENTS` array. One option is to change the value assigned to the `title` prop. A better option, I think, is to add an `allowFutureDates` prop to **`ModalDetails`** and set it to true where necessary. I've left a couple suggestions below on how the component would be updated.

src/components/ModalDetails/ModalDetails.js
spetros-do: updated. see comments below.

src/components/ModalDetails/ModalDetails.js
spetros-do: updated. see comments below.

spetros-do:
updated & reverted to a different PR: https://github.com/US-EPA-CAMD/easey-ecmps-ui/pull/1600

PR #1603, opened 2025-06-03
Fix/6677 checkout state

### LINKED ISSUE #6677: Checkout status not updating between opened tabs
When opening multiple tabs where one configuration is checked out and the other is not, the checkout state of the tab that is first opened for display will be used for the header information and it will not change when navigating to other opened facility tabs.

This affects the following controls in the specified modules:
- Monitoring Plans: **Import Data**, **Revert to Official Record**, & **Check Back In** buttons.
- Emissions: **Import Data** & **Check Back In** buttons.

## Steps to Reproduce

1. Navigate to the **Monitoring Plans** module, open a facility accordion, & "Open & Checkout" a configuration. Do not open the configuration's tab.
2. Open a different facility accordion & "Open" one of the facility's configurations.
3. Open the tab of the checked out plan. The controls noted above should be visible.
4. Open the tab of the plan that is _not_ checked out. The controls noted above will still be visible (incorrectly).
5. Go back to the **Select Configurations** tab.
6. Repeat steps 1-4, but open the not checked out location first. The controls will not be visible on the tab of the checked out configuration.

## Acceptance Criteria
- When multiple tabs are open in the Monitoring Plan module and the checked out configuration is displayed, the **Import Data**, **Revert to Official Record**, & **Check Back In** buttons should be visible.
- When multiple tabs are open in the Monitoring Plan module and a configuration that is not checked out is displayed, the **Import Data**, **Revert to Official Record**, & **Check Back In** buttons should _not_ be visible.
- When multiple tabs are open in the Emissions module and the checked out configuration is displayed, the **Import Data** & **Check Back In** buttons should be visible.
- When multiple tabs are open in the Emissions module and a configuration that is not checked out is displayed, the **Import Data** & **Check Back In** buttons should _not_ be visible.

maxdiebold-erg:
## Related Issues:

- https://github.com/US-EPA-CAMD/easey-ui/issues/6677

## Main Changes:

- Re-render the **`HeaderInfo`** component whenever the associated monitor plan ID changes. Since so much of the internal state depends on that ID, I decided to simply use the `key` prop to completely refresh the component rather than update the internal state.


EvanSun220 (approved):
Looking all good!

PR #1609, opened 2025-06-25
Add MP Comment Functionality

### LINKED ISSUE #6791: Add MP Comment Functionality
Continuation of #6740 

Add/Edit Monitoring Plan Comments screen from ECMPS 1.0:
Image

## Requirements
 
- For adding a new comment record, there should be three fields: Monitoring Plan Comment, Begin Date, and End Date
     - Monitoring Plan Comment and Begin Date fields are required when saving a new record.
          - If the Monitoring Plan Comment field is null, a MONPLAN-7-A screen check error should fire.
          - If the Begin Date field is null, a MONPLAN-4-A screen check error should fire
          - If the End Date is before the Begin Date, a MONPLAN-6-A screen check error should fire
- When a user saves a new MP comment, a record should be added in the CAMDECMPSWKS.MONITOR_PLAN_COMMENT table
     -  If a user saves a new comment record, the Monitoring Plan should be marked as resubmission required and will need to be evaluated.
- Users should be able to Edit MP comment records in the UI. Users can edit the Monitoring Plan Comment, Begin Date, and End Date fields
- Users should be able to remove Monitoring Plan Comments from the workspace that have not been submitted to official yet. 
- When a user submits a Monitoring Plan with a new/ edited comment, the comment should be loaded to the CAMDECMPS.MONITOR_PLAN_COMMENT table.
     - If a user tries to remove a MP comment that has already been submitted to official, an error message should fire. Error message from ECMPS 1.0 included below:
          - Image
- If a user imports a MP JSON with new comment records, the comments should appear in the UI.

## Acceptance Criteria

- Able to add new MP comment records in the UI and submit. After submission, revert to official and verify that comment is still present.
     - Verify that Monitoring Plan Comment and Begin Date fields are required to save a comment record.
- Able to edit MP comment records in the UI, save changes, and submit. After submission, revert to official and verify that edits to comment are still present.
     - Should be able to edit the Monitoring Plan Comment, Begin Date, and End Date fields.
- Able to import an MP JSON with new MP comment and view/ edit the comment in the UI
- Should be able to remove Monitor Plan Comments from the workspace that have not been submitted to official yet. 
- If a user tries to remove a MP comment that has already been submitted to official, and is also present in the workspace the comment should not be remove from official. 
- Able to evaluate changes in UI 
- Able to submit changes to official record in submission UI.

Jhernandez29:
Ticket:
https://github.com/US-EPA-CAMD/easey-ui/issues/6791
Changes: 
Create a new DataTableCommets table, added new modal input control "textarea" instead of input.  Added new column Monitor Plan Comment column and added wrap as true to show whole text. Deleted old comment view button and function. Added new getMonitoringPlansCommentsTableRecords function in monitoringPlanComments.js


src/additional-functions/extract-user-input.js
```diff
@@ -95,6 +95,29 @@ export const validateUserInput = (userInput, dataTableName) => {
...
+
+  const beginAfterEndMsg = "Begin date is after end date"
+  const commentError = "You did not provide comment, which is required for Monitor Plan Comment"
+  const { beginDate, beginHour, beginMinute, endDate, endHour, endMinute, monitoringPlanComment } = userInput
+
+  const isBeginBeforeEndDate = (beginDate && endDate)
+    ? checkBeginBeforeEndDate(beginDate, endDate, beginHour, endHour, beginMinute, endMinute)
+    : true
+
+  if (!isBeginBeforeEndDate) {
```
yonatan-dp: Missing validation check for beginDate? @Jhernandez29 Please make sure all parts of the requirement are captured. The ticket has details.

src/additional-functions/extract-user-input.js
```diff
@@ -95,6 +95,29 @@ export const validateUserInput = (userInput, dataTableName) => {
   return errors;
 };
 
+export const validateUserInputMonitorPlanComment = (userInput) => {
+  const errors = [];
+
+  const beginAfterEndMsg = "Begin date is after end date"
+  const commentError = "You did not provide comment, which is required for Monitor Plan Comment"
```
yonatan-dp: This is a generic error message.  Can you to see if there are specific required error messages for each validation. For instance,  MONPLAN-7-A check may have a standard error message. Same for the others.

src/components/datatablesContainer/DataTableComments/DataTableComments.js
```diff
@@ -0,0 +1,325 @@
...
+
+
+  const openCommentModal = (row, bool, create) => {
+    let comments = null;
+    setCreateNewComments(create);
+
+    if (commentData.length > 0 && !create) {
+      comments = commentData.filter((element) =>
+        {
+         return element.id === row[`col${Object.keys(row).length - 1}`]
```
yonatan-dp: This will break if we rearrange the column order. Can you try to get what you need without depending on the fact that the comment column is the last column in the table?

src/utils/api/monitoringPlansApi.js
```diff
@@ -268,6 +268,46 @@ export const saveMonitoringMats = async (payload) => {
   }
 };
 
+export const saveMonitorPlanComments = async (payload) => {
+
+  const url = getApiUrl(`/plans/${payload.planId}/comments/${payload.id}`);
+
+  // *** remove monitor plan comment not needed by the API
+  delete payload["planId"];
```
yonatan-dp: This is minor but it could introduce a bug in the future if someone moves the 'delete' lines above the URL construction. This is because we are changing the actual incoming parameter (which is not a good pattern).  Something like this would fix it (no need to delete)

 const { planId, id, ...apiPayload } = payload;
    const url = getApiUrl(`/plans/${planId}/comments/${id}`);

Then use apiPayload as the payload to for the api call

yonatan-dp (changes requested):

src/components/datatablesContainer/DataTableComments/DataTableComments.js
Jhernandez29: @yonatan-dp . This creates a new object array with records ->   data.forEach((el) => {
    records.push({
      col1: el?.monitoringPlanComment,
      col2: el?.beginDate,
      col3: el?.endDate,
      col4: el?.id
    });
  }). in monitorPlanComments.getMonitoringPlansCommentsTableRecords. This object will always have the id as the last value. Base on this I think my approach is correct.

src/additional-functions/extract-user-input.js
Jhernandez29: All validation checks are now done in the DTO.

src/utils/api/monitoringPlansApi.js
Jhernandez29: Put in place

src/additional-functions/extract-user-input.js
Jhernandez29: Doing all validation in monitoring plan API

yonatan-dp (approved):

PR #1709, opened 2025-08-27
508: Failed login "X" needs a title

### LINKED ISSUE #6834: 508: Failed login "X" needs a title
Validation error message box for a failed login includes an X button to dismiss the message. This button has no accessible name.

Not sure on this one as I AM seeing an aria-label... I suppose you could also add a title attribute with the same value.

Jhernandez29:
ticket:
https://github.com/US-EPA-CAMD/easey-ui/issues/6834

changes:
Updated Layout.js with IconButton to add the aria-label. This Component we don't need the onKeyUP anymore because is dynamically added already.

### DIFF
--- a/src/components/Layout/Layout.js
+++ b/src/components/Layout/Layout.js
@@ -15,8 +15,19 @@ import "./Layout.scss";
 import { hideAppError, hideAppWarning } from "../../additional-functions/app-error";
 
 import { getContent } from "../../utils/api/contentApi";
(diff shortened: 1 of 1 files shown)


ergjustin (approved):
Looks good to me.

- Confirmed that the aria-label="Close error alert" is present on the X (close) icon.
- Verified that the alert can be closed using the Enter key when the X icon is focused.

PR #1719, opened 2025-09-02
508: MP tabs

### LINKED ISSUE #6838: 508: MP tabs
Monitoring Plan: When Open or Open & Checkout is selected, a new view tab is added and focus correctly shifts to the newly added tab button.  The view does not refresh to the checked out configuration and BOTH tab selections remain actionable.  It will not be clear to low / no vision users which tab interface is being displayed.  Both buttons say they will open the tab but neither indicate which tab is already open. 

Decision: Take user to newly opened tab.

Include a state indicator in the aria label to indicate which tab is visible and active.

Jhernandez29:
ticket:
https://github.com/US-EPA-CAMD/easey-ui/issues/6838

changes:
Added aria-selected to take user to the newly opened tab. Updated aria-label with active to indicate which tab is visible and active.

### DIFF
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -100,9 +100,9 @@ const Tabs = ({
     tabBtnSelector = `[aria-label="${arg}"]`;
     return arg;
   };
-  const makeTabButtonAriaLabel = (pane) => {
+  const makeTabButtonAriaLabel = (pane, active) => {
     const parts = [
-      'open',
+      active,
       pane.title.split('(')[0].trim(),
       user &&
       pane.locationId &&
@@ -111,7 +111,7 @@ const Tabs = ({
       (isCheckedOut(pane.locationId) ||
         checkedOutLocations.some((loc) => loc.facId === parseInt(pane.facId)))
         ? '(locked)'
-        : '',
+        : '(not locked)',
       pane.title
         .split('(')[1]
         .replace(')', '')
@@ -120,7 +120,7 @@ const Tabs = ({
         .trim(),
       pane.locationId && isCheckedOutByUser(pane.locationId)
         ? '(checked-out)'
-        : '',
+        : '(not checked-out)',
       'tab',
     ].filter(Boolean);
 
@@ -163,9 +163,15 @@ const Tabs = ({
                       : 'tab-button react-transition flip-in-y'
                   }
                   tabIndex="0"
-                  aria-label={updateTabBtnSelectorAndReturnAriaLabel(
-                    makeTabButtonAriaLabel(pane),
+                  aria-label={
+                    updateTabBtnSelectorAndReturnAriaLabel(
+                    makeTabButtonAriaLabel(pane,((currentTabIndex === i)
+                      ? 'active'
+                      : ''))
                   )}
+                  aria-selected={((currentTabIndex === i)
+                      ? 'true'
+                      : 'false')}
                   onClick={() => {
                     addElementToLastFocusedArray(tabBtnSelector);
                     setCurrentTabIndex(i);
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -130,7 +130,7 @@ const Tabs = ({
   return (
     <div>
       <div className="tab-buttons mobile-lg:margin-left-7 mobile-lg:padding-left-5 tablet:margin-left-0 tablet:padding-left-0">
-        <ul className="usa-button-group margin-top-1">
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.js
```diff
@@ -100,9 +100,9 @@ const Tabs = ({
     tabBtnSelector = `[aria-label="${arg}"]`;
     return arg;
   };
-  const makeTabButtonAriaLabel = (pane) => {
+  const makeTabButtonAriaLabel = (pane, active) => {
     const parts = [
-      'open',
+      active,
```
maxdiebold-erg: The word "open" here is meant to be a verb representing the action that's executed when the button is clicked, not an adjective describing which tab is open. If we want to change it to be a descriptor instead, I think it should be parenthesized like the other state descriptors in the label. However, I'm not sure about using the word "active" in this case, since it already has a meaning in the context of monitoring plans. Maybe "(displayed)"?

maxdiebold-erg (changes requested):
The values for aria-selected and aria-label look good! I do have one suggestion below for the aria-label, though.

In addition to these changes, the requirements in the ticket state:

> Decision: Take user to newly opened tab.

This is not occurring in my testing.

Jhernandez29:
> The values for aria-selected and aria-label look good! I do have one suggestion below for the aria-label, though.
> 
> In addition to these changes, the requirements in the ticket state:
> 
> > Decision: Take user to newly opened tab.
> 
> This is not occurring in my testing.

@maxdiebold-erg 
If the screen reader opens the newly tab. The screen reader has to immediately read the open tab which aria-selected is doing this.  This is already happening.

src/components/Tabs/Tabs.js
Jhernandez29: I will change it to "displayed".

maxdiebold-erg:
I took that to mean that when "Open" or "Open & Checkout" is clicked, the tab is added **and** rendered on the screen.

src/components/Tabs/Tabs.js
Jhernandez29: After talking to @mark-hayward-erg. Aria-label doesn't have to change. 
We can just leave aria-selected

Jhernandez29:
> > The values for aria-selected and aria-label look good! I do have one suggestion below for the aria-label, though.
> > In addition to these changes, the requirements in the ticket state:
> > > Decision: Take user to newly opened tab.
> > 
> > 
> > This is not occurring in my testing.
> 
> @maxdiebold-erg If the screen reader opens the newly tab. The screen reader has to immediately read the open tab which aria-selected is doing this. This is already happening.

This is fix now

maxdiebold-erg (changes requested):
The new functionality works well! I did notice, though, that SonarQube flagged an error in the placement of the `aria-selected` attribute: "The attribute aria-selected is not supported by the role button." To resolve this, I think the buttons should be given the role "tab" instead, the enclosing `<ul>` should be given the role "tablist", and, optionally, the main tab content should be given the role "tabpanel".

I can't comment on unchanged lines, but here's a diff showing the changes I'd suggest:
```diff
diff --git a/src/components/Tabs/Tabs.js b/src/components/Tabs/Tabs.js
index f97a68c2..cc5d7020 100644
--- a/src/components/Tabs/Tabs.js
+++ b/src/components/Tabs/Tabs.js
@@ -130,7 +130,7 @@ const Tabs = ({
   return (
     <div>
       <div className="tab-buttons mobile-lg:margin-left-7 mobile-lg:padding-left-5 tablet:margin-left-0 tablet:padding-left-0">
-        <ul className="usa-button-group margin-top-1">
+        <ul className="usa-button-group margin-top-1" role="tablist">
           {panes.map((pane, i) => (
             <li
               key={i}
@@ -143,6 +143,7 @@ const Tabs = ({
                 <>
                   <Button
                     type="button"
+                    role="tab"
                     outline={currentTabIndex !== i}
                     tabIndex={0}
                     id="select-config"
@@ -158,7 +159,7 @@ const Tabs = ({
                 </>
               ) : (
                 <div
-                  role="button"
+                  role="tab"
                   id="tabBtn"
                   className={
                     currentTabIndex === i
@@ -248,7 +249,7 @@ const Tabs = ({
           ))}
         </ul>
       </div>
-      <div className="tabContent border-top-1px border-base-lighter margin-top-4 padding-top-4">
+      <div className="tabContent border-top-1px border-base-lighter margin-top-4 padding-top-4" role="tabpanel">
         {panes[currentTabIndex]?.content}
       </div>
     </div>
```

maxdiebold-erg:

maxdiebold-erg (approved):

PR #1736, opened 2025-09-11
QA Cert Preview focus

### LINKED ISSUE #6842: 508: QA Cert Preview focus
QA Certifications > Import Historical QA Test Data: When the Preview button is activated, focus should shift to the top of the new content (or maybe remain on the Preview button).  Instead, it shifts to the X icon to close the modal.

Focus should stay on the preview button.

Jhernandez29:
ticket:
https://github.com/US-EPA-CAMD/easey-ui/issues/6842

chages:
Added modal-title to make let screen readers know they are in  a modal. Focus <Button> from "@trussworks/react-uswds" right when the modal opens. 
Updated focus-trap.js logic with the desired element that we want by passing elementId to the Modal Component. Otherwise, always focus the first element.

### DIFF
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -107,7 +107,8 @@ export const Modal = ({
   const modalClassName = "modal-wrapper radius-md";
 
   return ReactDom.createPortal(
-    <div role="dialog" aria-modal="true">
+    <div role="dialog" aria-modal="true" aria-labelledby="">
+    <h2 id="modal-title">Preview Your Data</h2>
       {showDarkBg ? <div className="usa-overlay is-visible"></div> : null}
       <div ref={modalRef}>
         <modalContext.Provider value={{ close }}>
--- a/src/components/Modal/Modal.js
+++ b/src/components/Modal/Modal.js
@@ -23,6 +23,7 @@ export const Modal = ({
   left = "25%",
   returnFocus,
   save,
+  firstFocusElementById,
   saveStatus = DataStatus.IDLE,
   show,
   showCancel,
@@ -62,7 +63,7 @@ export const Modal = ({
   }, []);
 
   useEffect(() => {
-    const { handleKeyPress } = focusTrap(".modal-content", close);
+    const { handleKeyPress } = focusTrap(".modal-content", close, firstFocusElementById);
 
     // *** FOCUS TRAP
     document.addEventListener("keydown", handleKeyPress);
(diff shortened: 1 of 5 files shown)


src/components/Modal/Modal.js
```diff
@@ -107,7 +107,8 @@ export const Modal = ({
   const modalClassName = "modal-wrapper radius-md";
 
   return ReactDom.createPortal(
-    <div role="dialog" aria-modal="true">
+    <div role="dialog" aria-modal="true" aria-labelledby="">
```
maxdiebold-erg: Is the `h2` ID supposed to be here?
```suggestion
    <div role="dialog" aria-modal="true" aria-labelledby="modal-title">
```

src/components/QAImportHistoricalDataPreview/QAImportHistoricalDataPreview.js
```diff
@@ -161,12 +170,15 @@ export const QAImportHistoricalDataPreview = ({
...
         <div className="grid-col-fill padding-x-9 padding-top-3">
+          <div id="preview-button">
           <Button
+            tabIndex={0}
             className="width-card"
             onClick={() => fetchDataPreviewRecords()}
           >
             Preview
           </Button>
+          </div>
```
maxdiebold-erg: This works, but why not just assign the ID to the button?
```suggestion
          <Button
            tabIndex={0}
            className="width-card"
            onClick={() => fetchDataPreviewRecords()}
            id="preview-button"
          >
            Preview
          </Button>
```

src/components/QAImportHistoricalDataPreview/QAImportHistoricalDataPreview.js
```diff
@@ -54,6 +54,15 @@ export const QAImportHistoricalDataPreview = ({
...
   const userClicked = useRef(false);
 
+    useEffect(() => {
+      requestAnimationFrame(() => {
+     const wrapper = document.getElementById("preview-button");
+      if (wrapper) {
+        const realButton = wrapper.querySelector("button");
+        realButton?.focus();
+      }})
+    }, []);
```
maxdiebold-erg: This is not working as expected. It is being overridden in two places:

- `focus-trap.ts` line 58:
```javascript
  setTimeout(() => {
    firstComponentFocusableElement.focus();
  });
```

- `Modal.ts` line 75:
```javascript
    setTimeout(() => {
      if (document.querySelector("#closeModalBtn")) {
        document.querySelector("#closeModalBtn").focus();
      }
    });
```

I think, instead, the `Modal` component and `focusTrap` function should take an optional parameter, e.g. `primaryFocusableElementId`, which, if provided, is used for initial focus instead of the close button in these places. In that case, I think this effect can be removed.
```suggestion
```

maxdiebold-erg (changes requested):
Looks like the right idea, but the close button is still being focused in my testing. I've left a couple suggestions below.

src/components/QAImportHistoricalDataPreview/QAImportHistoricalDataPreview.js
Jhernandez29: Updated in last commit

src/components/Modal/Modal.js
Jhernandez29: updated in last commit

maxdiebold-erg (approved):
Works well!

PR #1751, opened 2025-09-29
fix/#6843 508 Add Buttons

### LINKED ISSUE #6843: 508: Add buttons
QA Certifications > Linearity Summary: The Test Data column header remains focusable but serves no function.  Because the Add button is inside the column header it may not be clear which focusable element functions as the Add button between the header and the actual button).  This issue also exists in the Linearity Test table and subsequent tables.

Maybe use the titles that are currently within the header and display as a caption instead, and then add the title into the Add button as an aria label.

Need to review all buttons in tables.

Image

### LINKED ISSUE #6844: 508: QA Add button tagging
QA Certifications > Linearity Summary: Test Data table and subsequent sub-tables include an Add button in the first column header.  The tagging is inconsistent with some of the buttons tagged simply as "Add".  These should be uniquely identifiable. 

Need more specificity for text/labels of Add buttons.

EvanSun220:
## Related Ticket:

- ticket [#6843](https://github.com/US-EPA-CAMD/easey-ui/issues/6843)
- https://github.com/US-EPA-CAMD/easey-ui/issues/6844

## Updates:

- Updated the first column header for the QAT, QCE and TEE screen.
- Updated the aria-label for all the Add buttons.
- Fixed the issue that user have to tab twice to focus on the Add button
- Add buttons have a uniquely identifiable


Jhernandez29 (approved):
Everything looks good.

mark-hayward-erg (approved):

PR #1794, opened 2025-12-04
#6939 Fix for this intermittent error: Unable to obtain user responsibilities from CBS

### LINKED ISSUE #6939: Unable to obtain user responsibilities from CBS
# Background

This error has been periodically popping up upon log in to ECMPS.  When I look at the Cloud.gov logs they indicate the when the auth-api called CBS for the user's permissions CBS responded with a 400 Bad Request.

Image

Looking at the CBS logs in Splunk you can see that occurring.

Image

I logged out and back in a few minutes later and everything was fine (also show in the above screen shot).  Why did the first request fail?

 - In these cases, the auth-mgmt/tokens/validate endpoint sometimes returns a response status code of 201, but also returns "false" as the response instead of the expected user information.   When CBS gets this "false" response, it is unable to retrieve the expected user information and the CBS responsibility API returns a 400 error.

# Changes/Updates Needed
 - (CBS) update the token validation logic in the responsibility API to check for the expected user object, and, if not found, return a 403 response with the message "Unauthorized: unable to validate user information associated with token." [This differs from the message returned when token validation is completed, but fails - that scenario returns a 403 response with the message "Unauthorized: invalid token.".]
 - (EASEY) update the response status code returned from auth-mgmt/tokens/validate to be [something other than 201, i.e., a  40x status code] when the user information cannot be found/returned.   
  - verify that this does not have an adverse impact on the ECMPS application itself
- (EASEY) when logging failures for calls to the CBS responsibility API, include the response data in the logs.  Currently, the logging for these failures only appears to include the response status code returned from CBS, but not the response data (which will include unique error messages that indicate the underlying issue for the failure).
 - possible responses from the CBS responsibility API
    | status code | response data | notes |
    | - | -| - |
    | 200              | '{"plantList":[{"id":"1","permissions":["DSMP"]},{"id":"2","permissions":["DSMP","DSQA"]}, "id":"3","permissions":[]}]}' | successful response |
    | 400              | '["Invalid parameters: missing userId."]' | userid is missing in the request |
    | 401              | '["Unauthorized: missing header."]' | clientIp is missing the the request header |
    | 403              | '["Unauthorized: invalid token."]' | token validation failed |
    | 403              | '["Unauthorized: unable to validate user information associated with token."]'  | token could not be validated because expected user information was not returned by auth-mgmt/tokens/validate |
    | 400              | '["Unable to get responsibilities."]' | error occurred retrieving responsibilities |
    | 400              | '["Unable to handle API call."]' | unexpected error occurred |
- (EASEY) investigate to determine why the user information cannot be found/returned condition occasionally occurs and make updates (if something concrete is found) to improve the behavior

# Acceptance Criteria Cases

## Case 1

> ### Steps:
> 1. Log into ECMPS 2.0 using a user id (originally DWHITTEN_SP).
> 2. Log out of ECMPS 2.0.
> 3. In the same browser and tab, log into ECMPS 2.0 using a different user id (originally DWHITTEN_AD).
>    - This user has the same Login.gov identity.
> 4. Login fails with the message "Login Error: Unable to obtain user responsibilities from CBS. Please try again later.".
> 
> ### Notes:
> 1. For Chrome
>    - Closing the browser and retrying step 3 does not clear the problem if another Chrome instance is open.
>    - Closing every Chrome instance and retrying step 3 does eliminate the problem.

yonatan-dp:
Fix for this intermittent error: Unable to obtain user responsibilities from CBS

### DIFF
--- a/src/components/Login/UserAccountStatus.js
+++ b/src/components/Login/UserAccountStatus.js
@@ -79,6 +79,7 @@ async function buildUrl(authUrl, policyResponse) {
     `p=${encodeURIComponent(p)}&` +
     `acr_values=${encodeURIComponent(acrValues)}&` +
     `scope=${encodeURIComponent(scope)}&` +
+    `prompt=login&` +
     `userId=${encodeURIComponent(policyResponse.userId.toUpperCase())}&` +
     `userRoleId=${encodeURIComponent(policyResponse.userRoleId)}`;
 }


ergjustin (approved):

yonatan-dp:
Closing this PR unmerged based on email exchange with the ICAM team.  

> From Christopher Chafin:
I don’t believe prompt=login will hurt anything during the signin flows. (It may break handoffs from MyCDX in CDX Web, or any other ICAM RSO/handoffs into the application). 

Closing to avoid any unintended side effects.

PR #1801, opened 2025-12-17
 Fix token refresh race condition causing authentication errors

### LINKED ISSUE #7018: Errors on Dev
On Dev: yesterday and today I keep on getting these errors (alternating back and forth) while actively working with ECMPS. I am then forced to logout and log back in before I can do anything else.

Image
Image

yonatan-dp:
Implemented atomic lock mechanism in refreshToken() to ensure only one request claims the refresh lock, and fixed catch block to return current token from localStorage instead of undefined. Undefined prevented self-healing on transient failures.

### DIFF
--- a/src/utils/api/easeyAuthApi.js
+++ b/src/utils/api/easeyAuthApi.js
@@ -173,15 +173,6 @@ export const refreshToken = async () => {
       localStorage.setItem("ecmps_refreshing_token", "false");
     }
 
-    let waitDuration = 0; //Keep track of how long the current call is occuring for, force it through after a certain amount of seconds
-    while (
-      localStorage.getItem("ecmps_refreshing_token") === "true" &&
-      waitDuration < config.app.tokenRefreshThresholdSeconds
-    ) {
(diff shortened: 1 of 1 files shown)


src/utils/api/easeyAuthApi.js
```diff
@@ -181,8 +181,8 @@ export const refreshToken = async () => {
       await new Promise((resolve) => setTimeout(resolve, 100));
       waitDuration += 100;
     }
```
maxdiebold-erg: I think this block is now redundant. If the current token is expired, the block below will wait for an ongoing call, but if not it will just use the existing token.

maxdiebold-erg (commented):
The changes look good! One question below about how they interact with pre-existing code.

src/utils/api/easeyAuthApi.js
yonatan-dp: That is correct!  I have removed that block.

maxdiebold-erg (approved):

PR #1802, opened 2025-12-17
QA Maintenance - date column sorting

### LINKED ISSUE #7020: QA Maintenance: date column sorting
On Dev: In QA Maintenance, the Begin Date & End Date columns sort as a string instead of as a date. This needs to be fixed as it makes it really hard to find the correct test.

Image

Jhernandez29:
ticket:
https://github.com/US-EPA-CAMD/easey-ui/issues/7020
changes:
Added sort function for react Table. This way they sort depending on the date and they are not treated as string.


ergjustin (approved):
Looks good to me

PR #1898, opened 2026-02-20
fix: added default value in the default table

### LINKED ISSUE #5980: Display the “Default Value” so users don’t have to open each record to see the value
Related to #6627. 

The Monitor Default grid does not display the actual default value (see screenshot). You have to click view/edit in order to see the value, but this shouldn't be happening.

### LINKED ISSUE #6627: User Feedback: Monitoring plan default record view enhancements
Related to #5980.

ECMPS 2.0 Default Record ER1

- Please display the emission rate on the main page so we don’t have to View/Edit to see the rate for the active/inactive record
- Possibly add an additional column where the highlighted arrow is
- I imagine room could be made by shrinking some of the other columns
- 

ECMPS 2.0 Default Record ER2 & ECMPS 2.0 Default Record ER3 - these are in the order they display when “View/Edit” is selected

- Increase the pop-up box size to display all of the data in the record as the first pop-up displays at the bottom of the box and has to be scrolled up to start at the top
- If the box size can not be increased, have the pop-up open at the top of the box
- The same applies to when I choose “Create Default”
- The Parameter needs to be chosen to enable the other fields
- No reason to start entering records at the bottom of the box

Image
Image
Image

ergjustin:
## Ticket
[Display the “Default Value” so users don’t have to open each record to see the value](https://github.com/US-EPA-CAMD/easey-ui/issues/5980)

## Changes

- Added default value in the default table


yonatan-dp (changes requested):
Does this change address all requirements on the ticket? Specifically: ECMPS 2.0 Default Record ER2 & ECMPS 2.0 Default Record ER3 - these are in the order they display when “View/Edit” is selected

I didn't run this locally but the code change seems to only address this requirement: ECMPS 2.0 Default Record ER1

ergjustin:
> Does this change address all requirements on the ticket? Specifically: ECMPS 2.0 Default Record ER2 & ECMPS 2.0 Default Record ER3 - these are in the order they display when “View/Edit” is selected
> 
> I didn't run this locally but the code change seems to only address this requirement: ECMPS 2.0 Default Record ER1

@yonatan-dp 
I believe this is already addressed. When selecting Edit or Create Default, the pop-up opens positioned at the top so the first values are visible without needing to scroll up.

Also, once the Parameter is selected, the dependent fields (Units of Measure, Purpose, Operating Condition, and Source of Value) become enabled as expected.
If anything is missing or you’re seeing different behavior, please let me know and I’ll fix it.
image

yonatan-dp (approved):
Looks good!

PR #1914, opened 2026-02-27
fix(css): Show column header text in full

### LINKED ISSUE #6916: Show column header text in full
Column headers should be shown in full in the entire app. Some places you can see the entire header, other places they are cut off with a "..." at the end.

A global update should be made to change the header rendering to wrap longer header text onto multiple lines.

More investigation needs to be done to flag all locations affected.

Acceptance Test:
- All the column headers in ECMPS 2.0 are entirely visible on the screen.

Developer Note:
Update _react-datatable-overrides.scss with

  .brFloq {
    white-space: pre-wrap !important;
    overflow: unset;
    text-overflow: unset;
  }

ergjustin:
## Ticket
[Show column header text in full ](https://github.com/US-EPA-CAMD/easey-ui/issues/6916)

## Changes 
Add css in _react-datatable-overrides.scss


public/scss/_react-datatable-overrides.scss
yonatan-dp, line 5: Should this updated also per the developer notes on the ticket (lines 5–7)?
 .data-display-table-report still has only white-space.

yonatan-dp (changes requested):

yonatan-dp (approved):
Looks good.

PR #1990, opened 2026-04-24
Fix/#7151 fix sub path routing

### LINKED ISSUE #7151: Fix Sub Path Routing in ECMPS
When `ecmps-ui` is mapped to a path-based route (e.g., `/ecmps`), the following issues occur:

**ecmps-ui**
1. `VITE_EASEY_ECMPS_UI_PATH` is not set in manifests, so `BrowserRouter basename` defaults to `/`.  Pages/scripts etc return 404 for all subpath requests
2. `index.html` and `error404.html` use hardcoded absolute paths (`/env-config.js`, `/favicon.ico`,
`/uswds/css/styles.css`), meaning they can't resolve from /ecmps. Results in 404
3. No `base` configured for VITE build output, so built JS/CSS bundles reference `/static/...` instead of relative paths
4. `error404.html` has hardcoded `href="/"` 

**auth-api**
1. ECMPS_UI_REDIRECT_URL needs to be updated to add /ecmps

### Acceptance Criteria

- [ ] App loads correctly at `<hostname>/ecmps` in all environments
- [ ] Static assets (JS, CSS, images, favicon) load without 404s when only the `/ecmps` path route exists (no root route)
- [ ] 404 error page renders with styles and links back to the correct home path
- [ ] ALL environments function ONLY on route `/ecmps` (e.g. on `https://ecmps-dev.app.cloud.gov/ecmps`,  **NOT on** `https://ecmps-dev.app.cloud.gov`

yonatan-dp:
Fix Sub Path Routing in ECMPS #7151


src/utils/api/easeyAuthApi.js
```diff
@@ -117,14 +117,17 @@ function storeUser(response) {
...
-    window.location.pathname.endsWith("/home") ||
-    window.location.pathname.endsWith("/")
+    pathWithoutBase.includes("/workspace") ||
+    pathWithoutBase.endsWith("/home") ||
+    pathWithoutBase.endsWith("/")
   ) {
     window.location.reload();
   } else {
-    window.location.assign(`/workspace${window.location.pathname}`);
+    window.location.assign(`${basePath}/workspace${pathWithoutBase}`);
```
djw4erg: SonarCloud produced a warning on using window and suggested using globalThis instead.  Not pushing for a change, but just noting.

djw4erg (dismissed):

src/utils/api/easeyAuthApi.js
yonatan-dp: Corrected these warnings. I think Sonar is going to flag this every time we touch those lines although they were pre-existing (and there are many more). But, I agree we should change them as we touch those lines even though the window.location references should work for ECMPS UI.

djw4erg (approved):

PR #2118, opened 2026-07-22
Feature/6037 bulk import

### LINKED ISSUE #6037: Add the ability to import multiple files at once for multiple locations (High-3)
#### Latest Import

The first screen when navigating to the "Bulk Import" tab will show the status of the latest bulk import for the user. This must draw from the last created row from an IMPORT_SET table, which will have child IMPORT_QUEUE records, similar to the evaluation submission tables. Each queue record will have it's own status and errors associated with it. The status message, for completed items, should be a link that, when clicked, opens a window with the generated report. The page should poll the status API to keep it up to date.

The table should have the following columns:

- File Name
- File Type _(MP, QA, or EM)_
- ORIS
- Unit/Stack/Pipe
- Status

On this page should be a "Start New Import" button that is disabled if the latest import is still in progress (i.e. only one bulk import is allowed at a time). This button shall direct to the "New Import" screen/modal.

#### New Import

This page will have a file input that populates a table with the following columns:

- File Name
- File Type
- Reporting Period _(EM only)_
- ORIS
- Unit/Stack/Pipe
- File Size (?)

An IMPORT_SET record should be created immediately with a DRAFT/NEW status. The files should be validated upon input, and sorted in the table by ORIS code, then by file type (MP -> QA -> EM), then by emissions quarter. The files should be uploaded to a folder in S3 keyed on the IMPORT_SET ID as they are input. Additionally, locations should be checked out when necessary as files are input. If a location cannot be checked out, show an error message and do not add the file to the list. The file input needs to be able to support adding and removing multiple files at a time.

Once "Submit" is clicked, a request to the `camd-services` API should be made to create IMPORT_QUEUE records, populating the `file_name` and `temp_s3_bucket_file_path` columns accordingly, and move the set's status to QUEUED. Depending on the result:

- Upon success, close the modal and display the import set status.
- Upon failure, clear the S3 files added in the previous step and show an error message.

### Locking

Two protective methods should be updated to avoid modifying related data while an import is in progress or queued. They need to be expanded to check for IMPORT_SET records that are in progress (will actually need to check the `mon_plan_id` associated with IMPORT_QUEUE records).

- **`UserCheckOutService.ensureNoEvaluationOrSubmissionInProgress`**
- **`RolesGuard.checkNotEvalOrSubmitted`**

### NOTE

A new S3 bucket needs to be created for this process. We should purpose it as a generic bucket for staging files.

maxdiebold-erg:
## Related Issues:

- https://github.com/US-EPA-CAMD/easey-ui/issues/6037

## Main Changes:

- Create the Bulk Import module, with a screen that shows the status of the latest submitted bulk import for the user, and a modal to select files for a new import.


src/components/BulkImport/BulkImport.jsx
```diff
@@ -0,0 +1,72 @@
...
+const POLL_MS = 10000;
+
+function BulkImport({ user }) {
+  const [latest, setLatest] = useState(null);
+  const [loading, setLoading] = useState(true);
+  const [showModal, setShowModal] = useState(false);
+
+  const loadLatest = useCallback(async () => {
+    const res = await getLatestImport();
+    console.log(res?.data);
```
ergjustin: I think a console.log was accidentally left in after debugging. Could you please remove it?

src/components/BulkImport/NewImportModal.jsx
```diff
@@ -0,0 +1,212 @@
...
+      exitBtn="Submit"
+      close={cancel}
+      save={submit}
+      saveStatus={submitting ? DataStatus.PENDING : DataStatus.IDLE}
+      errorMsgs={errorMsgs}
+      disableExitBtn={processing || files.length === 0}
+      width="75%"
+    >
+      <div className="padding-top-2">
+        <input
```
ergjustin: For the UI, when I click Import, it currently shows the default Choose Files input. Would it be possible to make the New Import modal look similar to the Monitoring Import (or the other import screens) so the UI stays consistent?

ergjustin (changes requested):
Looks good to me! I’m able to access and see the workspace/bulk-import page.
Just a couple of small suggestions:

src/components/BulkImport/BulkImport.jsx
maxdiebold-erg: Good catch!

src/components/BulkImport/NewImportModal.jsx
maxdiebold-erg: I will look into it. The problem is that the existing component doesn't allow selecting multiple files, so will need to update that or create a new one, and figure out how multiple files should display.

maxdiebold-erg: Should be using the common component, now, looks like the component did already accept multiple files.

ergjustin (approved):
Thanks for the changes.
Looks good to me.

